PR #2, opened 2023-09-16
Feat/decouple and add tests

Eikix:
I propose in this PR to decouple Relocatable and Felt
Let me know if you think that's "clean"

In our case, it'd make us stray away from OOP in some way: we cannot access MaybeRelocatable as if it were a class. But this limits redundancy and both Felt and Relocatable don't have to know about each other.


src/memory/types.ts
ClementWalter, line 1: why do we put errors in a dedicated `types.ts`? why not `errors.ts` then?

src/memory/primitives/felt.ts
```diff
@@ -0,0 +1,41 @@
...
+  add(other: Felt): Felt {
+    return new Felt(this.inner + other.inner);
+  }
+
+  sub(other: Felt): Felt {
+    return new Felt(this.inner - other.inner);
+  }
+
+  getInner(): BigInt {
+    return this.inner;
```
ClementWalter, line 21: this is contradictory with the fact that it's `private`, why would we want this?

src/memory/primitives/felt.test.ts
```diff
@@ -0,0 +1,38 @@
+import { test, expect, describe } from 'bun:test';
+import { ConversionError, Felt } from './felt';
+
+describe('felt', () => {
+  test('should initialise a felt correctly and perform good conversions', () => {
+    const felt = new Felt(10n);
+
+    expect(felt.getInner()).toEqual(10n);
```
ClementWalter, line 8: maybe we should implement a `.eq` then instead of this `getInner`

src/memory/primitives/felt.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    expect(felt.toNumber()).toEqual(10);
+    expect(felt.toString()).toEqual('10');
+    expect(felt.toHexString()).toEqual('a');
```
ClementWalter, line 11: use one assert per test

src/memory/primitives/felt.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  test('should fail number conversion when felt inner > JS max number', () => {
+    // twoPowHundred == 2^100
+    const twoPowHundred = 1267650600228229401496703205376n;
```
ClementWalter, line 16: you can just write `2n ** 100n` and remove the comment

ClementWalter: and actually the test should use Number.MAX_SAFE + 1 for example, and not a random pow (here 100)

src/memory/primitives/felt.test.ts
```diff
@@ -0,0 +1,38 @@
+import { test, expect, describe } from 'bun:test';
+import { ConversionError, Felt } from './felt';
+
+describe('felt', () => {
```
ClementWalter, line 4: add an extra level of `describe` for the function name

```
describe("Felt"), () => {
    desribe("add", () => {

    })
    describe("sub", () => {

    })
    // etc.
}
```

src/memory/primitives/maybeRelocatable.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import { Felt } from './felt';
+import {
+  ForbiddenOperation,
+  OffsetUnderflow,
+  Relocatable,
+} from './relocatable';
+import { add, sub } from './maybeRelocatable';
+import { InternalError } from '../types';
+
+describe('maybe relocatable', () => {
```
ClementWalter, line 11: describe should describe the System Under Testint (SUT), here the type `MaybeRelocatable`, hende 

```
describe("MaybeRelocatable")
```

src/memory/primitives/maybeRelocatable.test.ts
ClementWalter, line 1: see comments in `felt.test.ts` regarding tests organization

src/memory/primitives/relocatable.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    if (this.segmentIndex !== other.segmentIndex) {
+      throw new SegmentError();
+    }
+
+    return new Relocatable(this.segmentIndex, this.offset - other.offset);
+  }
+
+  getSegmentIndex(): number {
+    return this.segmentIndex;
```
ClementWalter, line 32: for a `Relocatable`, maybe we should put `segmentIndex` and `offset` as `public` and avoid these getters

src/memory/primitives/relocatable.test.ts
ClementWalter, line 1: see comments in other test files

ClementWalter (changes requested):

ClementWalter (changes requested):
I wondering if it's really better to define the `MaybeRelocatable` like that, instead of just importing the type in `Felt` and `Relocatable`, it seems like it creates a lot of code

ClementWalter (changes requested):
I wondering if it's really better to define the `MaybeRelocatable` like that, instead of just importing the type in `Felt` and `Relocatable`, it seems like it creates a lot of code

src/memory/primitives/felt.test.ts
Eikix: we'd still need get getInner to perform the equality check 
```
eq(other: Felt): boolean {
    return this.inner == other._getInner();
  }
```

src/memory/primitives/felt.test.ts
Eikix: ok

src/memory/primitives/relocatable.ts
Eikix: then we expose the possibility to do `someRelocatable.segmentIndex += 1` in some code, instead of using setter functions

Eikix:
> I wondering if it's really better to define the `MaybeRelocatable` like that, instead of just importing the type in `Felt` and `Relocatable`, it seems like it creates a lot of code

Good point, my point was to decouple Felt class from everything since I sensed we might need to switch which felt type we'd use -> Lambda VM go uses a Felt implementation written in lambdaworks

Maybe they used a specific felt implem. so that:
- everything along the stack would use the same felt type -> vm, blockifier in go, and then application
- for performance -> their felt type is highly optimized


I propose a refactor ->

Relocatable is allowed to know about Felt,
Felt is not allowed to know about Relocatable

Eikix:
Opened #4 instead

PR #4, opened 2023-09-20
Feat/add prime arithmetics and tests

Eikix:
Refactored the whole PR to respect comments


src/memory/primitives/relocatable.ts
```diff
@@ -0,0 +1,80 @@
...
+      if (this.getOffset() + other.toNumber() > Number.MAX_SAFE_INTEGER) {
+        throw new OffsetOverflow();
+      }
+      return new Relocatable(
+        this.getSegmentIndex(),
+        this.getOffset() + other.toNumber()
+      );
+    }
+
+    if (other instanceof Relocatable) {
```
greged93, line 42: Do you need this check? If it's not Felt it should be Relocatable no?

src/memory/primitives/relocatable.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    test('should subtract a Felt wrapped in MaybeRelocatable', () => {
+      const a = new Relocatable(0, 10);
+      const b = new Felt(5n);
+      const result = a.sub(b);
+      expect(result.getOffset()).toEqual(5);
+    });
+  });
+
+  describe('add', () => {
+    test('should add a Felt wrapped in MaybeRelocatable', () => {
```
greged93, line 68: I think you forgot to wrap in a MaybeRelocatable

src/memory/primitives/relocatable.test.ts
```diff
@@ -0,0 +1,99 @@
...
+import { Felt } from './felt';
+import {
+  Relocatable,
+  OffsetUnderflow,
+  SegmentError,
+  ForbiddenOperation,
+} from './relocatable';
+import { UnsignedInteger } from './uint';
+
+describe('Relocatable', () => {
```
greged93, line 11: For most of these tests we assert that the offset moved, should we also assert that the segment didn't move?

src/memory/primitives/relocatable.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      const result = relocatable.add(felt);
+      expect(result.getOffset()).toEqual(10);
+    });
+    test('should add a positive number correctly to a relocatable', () => {
+      const relocatable = new Relocatable(0, 5);
+      const result = relocatable.add(UnsignedInteger.toUint(5));
+      expect(result.getOffset()).toEqual(10);
+    });
+
+    test('should throw when adding a negative number', () => {
```
greged93, line 92: This test should be in uint.test.ts, you are throwing because `UnsignedInteger.toUint(-5)` will throw.

src/memory/primitives/relocatable.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      const relocatable = new Relocatable(0, 10);
+      const felt = new Felt(5n);
+      const result = relocatable.sub(felt);
+      expect(result.getOffset()).toEqual(5);
+    });
+
+    test('should throw OffsetUnderflow when subtracting a larger Felt', () => {
+      const relocatable = new Relocatable(0, 5);
+      const felt = new Felt(10n);
+      expect(() => relocatable.sub(felt)).toThrow(new OffsetUnderflow());
```
greged93, line 49: in `uint.test.ts` you are using this syntax for checking that a test throws:
```
expect(() => UnsignedInteger.toUint(-5)).toThrow(TypeError);
```
can we sync on one syntax?

greged93 (commented):
couple of comments, very nice though, I'm almost starting to like ts

src/memory/primitives/relocatable.test.ts
Eikix: ah yeah no need to wrap actually i'll change the text

src/memory/primitives/relocatable.test.ts
Eikix: mhm good question, ok why not

src/memory/primitives/relocatable.test.ts
greged93: But then you can remove it, I think you have the same test 15 lines lower

src/memory/primitives/relocatable.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export class RelocatableError extends Error {}
+export class OffsetUnderflow extends RelocatableError {}
+export class OffsetOverflow extends RelocatableError {}
+export class SegmentError extends RelocatableError {}
+export class ForbiddenOperation extends RelocatableError {}
+export class TypeError extends RelocatableError {}
+export class InternalError extends RelocatableError {}
+
+export type MaybeRelocatable = Relocatable | Felt;
```
greged93, line 12: The structure is a little weird no? Because you implement some operations between Relocatables and MaybeRelocatables. I would keep it the cairo-vm-go way:
addition with `add` is only allowed between the same types. If you want to add a Felt to a MaybeRelocatable for example, I would explicitly have a `addFelt` function.

I also think we should maybe just implement addition between types, and `addFelt` (I think this is what lambda class did).

src/memory/primitives/relocatable.ts
Eikix: it can be a Uint

greged93 (approved):
lgtm

PR #6, opened 2023-09-23
fix: memory management

greged93:
Fixes couple of issues in the memory, typos and adds couple of tests for the Memory.

### DIFF
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -9,13 +9,13 @@ export class MemoryError extends Error {}
 export class WriteOnceError extends MemoryError {}
 export class UnknownAddressError extends MemoryError {}
 
-class Memory {
+export class Memory {
   data: Map<Relocatable, MaybeRelocatable>;
   numSegments: Uint;
 
-  constructor(data: Map<Relocatable, MaybeRelocatable>, numSegments: number) {
-    this.data = data;
-    this.numSegments = UnsignedInteger.toUint(numSegments);
+  constructor() {
+    this.data = new Map();
+    this.numSegments = UnsignedInteger.toUint(0);
   }
 
   insert(address: Relocatable, value: MaybeRelocatable) {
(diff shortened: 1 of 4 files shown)


src/memory/memory.ts
```diff
@@ -30,37 +30,36 @@ class Memory {
...
+export class MemorySegmentManager {
+  segmentSizes: Record<Uint, Uint>;
   memory: Memory;
 
-  constructor(segmentSizes: Record<number, number>, memory: Memory) {
-    this.segmentSizes = segmentSizes;
-    this.memory = memory;
+  constructor() {
+    this.segmentSizes = {};
+    this.memory = new Memory();
```
Eikix, line 48: I think we'd want to be able to pass an existing memory, so as to wrap it?
Especially if we use it at the end of a program

src/memory/memory.ts
```diff
@@ -30,37 +30,36 @@ class Memory {
...
       throw new UnknownAddressError();
     }
+    return value;
   }
 }
 
-class MemorySegmentManager {
-  segmentSizes: Record<number, number>;
+export class MemorySegmentManager {
+  segmentSizes: Record<Uint, Uint>;
```
Eikix, line 43:
```
In general:

If you have a static set of keys or you want a type that behaves like a regular object but with type safety, Record is a good choice.
If you want a dynamic collection of key-value pairs with more utility methods and insertion order, then Map is the way to go.
Lastly, considering the branding on Uint, if you were to use the Record, keep in mind that object keys in JavaScript are always strings, so there might be some nuances with key conversion if you try to use the branded number directly as a key.
```

src/memory/memory.ts
```diff
@@ -30,37 +30,36 @@ class Memory {
...
   }
 
-  loadData(address: Relocatable, data: MaybeRelocatable[]) {
+  loadData(address: Relocatable, data: MaybeRelocatable[]): Relocatable {
     data.forEach((d, index) =>
-      this.memory.insert(
-        address.addPositiveNumber(UnsignedInteger.toUint(index)),
-        d
-      )
+      this.memory.insert(address.add(UnsignedInteger.toUint(index)), d)
```
Eikix, line 61: Did we check for exceptions? for instance, we cannot upsert

src/memory/primitives/felt.test.ts
```diff
@@ -3,10 +3,10 @@ import { ConversionError, Felt, FeltError } from './felt';
 
 describe('Felt', () => {
   describe('constructor', () => {
-    test('should throw if initialising a felt with a negative inner', () => {
+    test('should throw if initializing a felt with a negative inner', () => {
```
Eikix, line 6: grrr why r we using american keyboard?

src/memory/primitives/felt.ts
```diff
@@ -41,7 +41,7 @@ export class Felt {
   }
 
   toNumber(): number {
-    let num = Number(this.inner);
+    const num = Number(this.inner);
```
Eikix, line 44: nice catch

Eikix (approved):

PR #8, opened 2023-09-23
test: segment manager

greged93:
Adds simple tests for the memory manager and splits it from the memory.ts file.

### DIFF
--- a/src/memory/memoryManager.test.ts
+++ b/src/memory/memoryManager.test.ts
@@ -1,0 +1,39 @@
+import { describe, test, expect } from 'bun:test';
+import { MemorySegmentManager } from './memoryManager';
+import { Relocatable } from './primitives/relocatable';
+import { Felt } from './primitives/felt';
+
+describe('MemoryManager', () => {
+  describe('addSegment', () => {
+    test('should add a new segment to the memory and return a pointer to it', () => {
+      const memoryManager = new MemorySegmentManager();
+      const segment = memoryManager.addSegment();
+
+      expect(segment).toEqual(new Relocatable(0, 0));
+    });
+    test('should expand the memory size', () => {
+      const memoryManager = new MemorySegmentManager();
+      const segment = memoryManager.addSegment();
+
+      expect(memoryManager.memory.numSegments).toEqual(1);
+    });
+  });
+  describe('loadData', () => {
+    test('should load data into the memory', () => {
+      const memoryManager = new MemorySegmentManager();
+      memoryManager.addSegment();
+      const data = [
+        new Relocatable(0, 0),
+        new Relocatable(0, 1),
+        new Felt(1n),
+        new Felt(2n),
+        new Relocatable(1, 1),
+      ];
+      const address = new Relocatable(0, 0);
+
+      expect(memoryManager.loadData(address, data)).toEqual(
+        new Relocatable(0, 5)
+      );
+    });
+  });
+});
--- a/src/memory/memoryManager.test.ts
+++ b/src/memory/memoryManager.test.ts
@@ -31,7 +31,7 @@ describe('MemoryManager', () => {
       ];
       const address = new Relocatable(0, 0);
 
-      expect(memoryManager.loadData(address, data)).toEqual(
+      expect(memoryManager.loadData(address, data).unwrap()).toEqual(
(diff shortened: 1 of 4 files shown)


src/memory/memoryManager.ts
```diff
@@ -0,0 +1,28 @@
...
+export class MemorySegmentManager {
+  segmentSizes: Map<Uint, Uint>;
+  memory: Memory;
+
+  constructor() {
+    this.segmentSizes = new Map();
+    this.memory = new Memory();
+  }
+
+  addSegment(): Relocatable {
```
Eikix, line 14: what i noticed is that in memory.test.ts we access directly numSegments
should it be private?

src/memory/memoryManager.ts
```diff
@@ -0,0 +1,28 @@
...
+    const ptr = new Relocatable(this.memory.numSegments, 0);
+    this.memory.numSegments = UnsignedInteger.toUint(
+      this.memory.numSegments + 1
+    );
+    return ptr;
+  }
+
+  loadData(address: Relocatable, data: MaybeRelocatable[]): Relocatable {
+    data.forEach((d, index) =>
```
Eikix, line 23: i had to modify the forEach into a for loop to accomodate result based

src/memory/memoryManager.test.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Relocatable } from './primitives/relocatable';
+import { Felt } from './primitives/felt';
+
+describe('MemoryManager', () => {
+  describe('addSegment', () => {
+    test('should add a new segment to the memory and return a pointer to it', () => {
+      const memoryManager = new MemorySegmentManager();
+      const segment = memoryManager.addSegment();
+
+      expect(segment).toEqual(new Relocatable(0, 0));
```
Eikix, line 12: should we assert on the numSegment?

src/memory/memoryManager.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    });
+    test('should expand the memory size', () => {
+      const memoryManager = new MemorySegmentManager();
+      const segment = memoryManager.addSegment();
+
+      expect(memoryManager.memory.numSegments).toEqual(1);
+    });
+  });
+  describe('loadData', () => {
+    test('should load data into the memory', () => {
```
Eikix, line 22: should we add a test to check that the memory was written to?

Eikix (commented):
lgmt

src/memory/memoryManager.ts
greged93: It's probably safer indeed, will see what I can do

src/memory/memoryManager.ts
greged93: let's merge your PR first then, I will rebase

src/memory/memoryManager.ts
greged93: Should I make a setter then? Because we need to update these fields from the memory manager

greged93:
Can we merge this @Eikix ?

src/memory/memoryManager.ts
```diff
@@ -32,6 +32,14 @@ export class MemorySegmentManager {
...
-    return address.add(UnsignedInteger.toUint64(BigInt(data.length)));
+
+    const segmentSize = UnsignedInteger.toUint64(BigInt(data.length));
+    this.segmentSizes.set(address.getSegmentIndex(), segmentSize);
+
+    return address.add(segmentSize);
+  }
+
+  getSegmentSize(segmentIndex: Uint64): Uint64 {
+    return this.segmentSizes.get(segmentIndex) || UnsignedInteger.toUint64(0n);
```
Eikix, line 43: i recommend operator `??` rather than || 

```
const a = null;
const b = undefined;
const c = "";
const d = 0;
const e = false;

console.log(a || "default");  // "default"
console.log(b || "default");  // "default"
console.log(c || "default");  // "default"
console.log(d || "default");  // "default"
console.log(e || "default");  // "default"

console.log(a ?? "default");  // "default"
console.log(b ?? "default");  // "default"
console.log(c ?? "default");  // ""
console.log(d ?? "default");  // 0
console.log(e ?? "default");  // false
```

Eikix (approved):

PR #10, opened 2023-09-24
Feat/runcontext

Eikix:


Eikix:
Proposed a big refactor of the result pattern, in order to gain more protection from the typescript compiler!

Before, when writing
```
if (num.isErr()) {
    // TypeScript knows it's Result<never, E>
} else {
    // TypeScript DOES NOT know it's Result<T, never>
}
```

Now:
if (num.isErr()) {
    // TypeScript knows it's Result<never, E>
} else {
    // TypeScript knows it's Result<T, never>
}

src/vm/virtualMachine.ts
```diff
@@ -0,0 +1,46 @@
...
+      return maybeEncodedInstruction;
+    }
+
+    const encodedInstruction = maybeEncodedInstruction.unwrap();
+
+    if (!(encodedInstruction instanceof Felt)) {
+      return new Err(InstructionError);
+    }
+
+    const maybeUint = encodedInstruction.toUint();
```
greged93, line 37: I don't think we need to call toUint, I think the instruction needs to be limited to Uint64

Eikix: problem is that JS's max integer is 2^53 so that's like U53? 
We need bigint for those 11 last bits,

Eikix: I could create a U64 type though like Uint, but with a higher bound

src/vm/instruction.ts
```diff
@@ -0,0 +1,66 @@
...
+// If the flag == 0, then the offset will point to Ap
+type ApRegisterFlag = 0;
+// If the flag == 1, then the offset will point to Fp
+type FpRegisterFlag = 1;
+type RegisterFlag = ApRegisterFlag | FpRegisterFlag;
+type DstRegister = RegisterFlag;
+type Op0Register = RegisterFlag;
+
+// Op1Src
+type Op1Src = 0 | 1 | 2 | 4;
```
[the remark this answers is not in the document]
Eikix: Not really you're right

src/vm/virtualMachine.ts
greged93: Would some helpers on Uint work? We could just have something like `isUint64`?

src/memory/memory.ts
```diff
@@ -50,7 +50,7 @@ export class Memory {
 }
 
 export class MemorySegmentManager {
-  segmentSizes: Record<Uint, Uint>;
+  segmentSizes: Record<number, Uint>;
```
greged93, line 53: Any reason to change this? I don't think the segments index can be negative, so having Uint makes sense no?

src/vm/instruction.ts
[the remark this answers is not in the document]
greged93: I think op1 src can also be an enum:
https://github.com/NethermindEth/cairo-vm-go/blob/main/pkg/vm/instruction.go#L28

src/vm/instruction.ts
```diff
@@ -0,0 +1,66 @@
...
+
+// ResLogic
+type ResLogic = 0 | 1 | 2 | 3;
```
greged93, line 34: Same for res logic I believe:
https://github.com/NethermindEth/cairo-vm-go/blob/main/pkg/vm/instruction.go#L52

src/memory/memory.ts
Eikix: I'm not allowed to have a non-native type as key, so it has to be number or string,
since former Uint was number wrapped with addons, it was fine, now that its a bigint , i cannot

src/vm/virtualMachine.ts
Eikix: Mmhm maybe but we'd need a new type for typeguards to truly be useful

greged93 (approved):
lgtm

PR #13, opened 2023-09-29
feat: uint

### LINKED ISSUE #11: dev: generic Uint<bits> type
Make a generic Uint type that would take as input a number of bits. The backend of the type would then use either `number` or `bigint` based on the amount of bits (`number` if 0 < bits < 53, otherwise `bigint`).

greged93:
Implement a generic type for Uint. Currently only use Uint64 and Uint32, but can be updated to have Uint8, Uint16,...
Can also be used for Int8, Int16,...
Resolves #11

### DIFF
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -24,7 +24,7 @@ export class Memory {
 
   constructor() {
     this.data = new Map();
-    this.numSegments = UnsignedInteger.toUint32(0);
+    this.numSegments = new UnsignedInteger().ZERO;
   }
 
   insert(address: Relocatable, value: MaybeRelocatable): Result<true, VMError> {
@@ -49,7 +49,13 @@ export class Memory {
   }
 
   incrementNumSegments() {
-    this.numSegments = UnsignedInteger.toUint32(this.numSegments + 1);
+    const newNumSegments = UnsignedInteger.toUint32(this.numSegments + 1);
+    if (newNumSegments.isErr()) {
+      throw new MemoryError(
+        'MemoryError: error incrementing number of segments'
+      );
+    }
+    this.numSegments = newNumSegments.unwrap();
   }
 
   getNumSegments(): Uint32 {
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -24,7 +24,7 @@ export class Memory {
 
   constructor() {
     this.data = new Map();
-    this.numSegments = new UnsignedInteger().ZERO;
+    this.numSegments = UnsignedInteger.ZERO;
   }
 
   insert(address: Relocatable, value: MaybeRelocatable): Result<true, VMError> {
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -4,7 +4,7 @@ import {
   Relocatable,
   SegmentError,
 } from 'primitives/relocatable';
-import { Uint64, UnsignedInteger } from 'primitives/uint';
+import { Uint32, UnsignedInteger } from 'primitives/uint';
 
 export class MemoryError extends Error {}
 
@@ -15,16 +15,16 @@ export const UnknownAddressError = {
 
 export const WriteOnceError = {
   message:
-    'MemoryError: tried to write exisiting memory. Can only write to memory once.',
+    'MemoryError: tried to write existing memory. Can only write to memory once.',
 };
 
 export class Memory {
   data: Map<Relocatable, MaybeRelocatable>;
-  private numSegments: Uint64;
+  private numSegments: Uint32;
 
   constructor() {
     this.data = new Map();
-    this.numSegments = UnsignedInteger.toUint64(0n);
+    this.numSegments = UnsignedInteger.toUint32(0);
   }
 
   insert(address: Relocatable, value: MaybeRelocatable): Result<true, VMError> {
@@ -49,10 +49,10 @@ export class Memory {
   }
 
   incrementNumSegments() {
-    this.numSegments = UnsignedInteger.toUint64(this.numSegments + 1n);
+    this.numSegments = UnsignedInteger.toUint32(this.numSegments + 1);
   }
 
-  getNumSegments(): Uint64 {
+  getNumSegments(): Uint32 {
     return this.numSegments;
   }
 }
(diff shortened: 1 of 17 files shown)


src/memory/memoryManager.ts
```diff
@@ -23,7 +23,7 @@ export class MemorySegmentManager {
     data: MaybeRelocatable[]
   ): Result<Relocatable, VMError> {
     for (let index = 0; index < data.length; index++) {
-      const sum = address.add(UnsignedInteger.toUint64(BigInt(index)));
+      const sum = address.add(UnsignedInteger.toUint32(index));
```
Eikix, line 26:
```suggestion
      const next_address = address.add(UnsignedInteger.toUint32(index));
```

src/memory/memoryManager.ts
```diff
@@ -33,13 +33,13 @@ export class MemorySegmentManager {
       }
     }
 
-    const segmentSize = UnsignedInteger.toUint64(BigInt(data.length));
+    const segmentSize = UnsignedInteger.toUint32(data.length);
```
Eikix, line 36: I think the segment size should be the sum of data.length and something,

Possibly: address.getOffset() + data.length
But what about memory holes?

src/primitives/felt.ts
```diff
@@ -43,6 +43,14 @@ export class Felt {
     return this.inner.toString();
   }
 
+  toUint32(): Result<Uint32, VMError> {
+    if (this.inner < 0n) {
```
Eikix, line 47: I think we should check here that the `inner` is below `Number.MAX_SAFE_INTEGER` as otherwise, toUint32 will panic

src/primitives/uint.ts
```diff
@@ -12,10 +27,28 @@ export class UnsignedInteger {
...
+
+  static upCast(num: Uint32): Uint64 {
+    return BigInt(num) as Uint64;
+  }
+
+  static downCast(num: Uint64): Uint32 {
+    if (num > 0x100000000n) {
+      throw new TypeError();
+    }
+    return Number(num) as Uint32;
```
Eikix, line 52:
```suggestion
    return num as Uint32;
```

I think should work

src/primitives/uint.ts
```diff
@@ -12,10 +27,28 @@ export class UnsignedInteger {
...
       return num;
     }
     throw new TypeError();
   }
+
+  static upCast(num: Uint32): Uint64 {
+    return BigInt(num) as Uint64;
+  }
+
+  static downCast(num: Uint64): Uint32 {
```
Eikix, line 48: do we need this?

src/primitives/uint.ts
```diff
@@ -12,10 +27,28 @@ export class UnsignedInteger {
...
+  }
+
   static toUint64(num: bigint): Uint64 {
     if (this.isUint64(num)) {
       return num;
     }
     throw new TypeError();
   }
+
+  static upCast(num: Uint32): Uint64 {
```
Eikix, line 44: do we need this?

src/memory/memoryManager.ts
greged93: I don't think memory holes are relevant here right?

src/primitives/felt.ts
greged93: I updated this previously I think, I can actually also remove the `this.inner < 0n`. My opinion is that this check is already done in the `UnsignedInteger.toUint32`. I can however have it so that `UnsignedInteger.toUint32` doesn't throw but returns an error, what do you think?

src/primitives/uint.ts
greged93: This doesn't work because in the case `num` is a `bigint` so I need to use the Number constructor.

src/primitives/uint.ts
greged93: Not right now no, can remove if you prefer

src/primitives/felt.ts
Eikix: Good idea

src/primitives/uint.ts
Eikix: Oh but if you're sure we'll need it why not

src/primitives/uint.ts
Eikix: Ah ok, then lgtm

src/primitives/uint.ts
```diff
@@ -1,8 +1,43 @@
...
+    'NumberConversionError: cannot convert to number, as underlying bigint < 0 or > Number.MAX_SAFE_INTEGER',
+};
+
+export const Uint64ConversionError = {
+  message:
+    'Uint64ConversionError: cannot convert to Uint64, as underlying bigint < 0 or > 2^64',
+};
 
 export class UnsignedInteger {
+  readonly ZERO: Uint32 = 0 as Uint32;
```
Eikix, line 30:
```suggestion
  static readonly ZERO: Uint32 = 0 as Uint32;
```

src/run-context/runContext.test.ts
```diff
@@ -6,11 +6,11 @@ describe('RunContext', () => {
   describe('incrementPc', () => {
     test('should successfully increment pc', () => {
       const ctx = new RunContext();
-      const instructionSize = UnsignedInteger.toUint64(2n);
-      const result = ctx.incrementPc(instructionSize);
+      const instructionSize = UnsignedInteger.toUint32(2).unwrap();
```
Eikix, line 9: how come we can call unwrap() here?

src/memory/memory.ts
```diff
@@ -15,16 +15,16 @@ export const UnknownAddressError = {
...
 
 export class Memory {
   data: Map<Relocatable, MaybeRelocatable>;
-  private numSegments: Uint64;
+  private numSegments: Uint32;
 
   constructor() {
     this.data = new Map();
-    this.numSegments = UnsignedInteger.toUint64(0n);
+    this.numSegments = new UnsignedInteger().ZERO;
```
Eikix, line 27:
```suggestion
    this.numSegments = UnsignedInteger.ZERO;
```

src/run-context/runContext.test.ts
greged93: don't think we can, I don't know why this passed

Eikix (approved):
lgtm

PR #15, opened 2023-10-16
feat: decode instruction

### LINKED ISSUE #12: feat: implement instruction decoding
Decode a cairo instruction. Source and helpers:
- https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/vm/decoding/decoder.rs#L12
- https://github.com/lambdaclass/cairo-vm_in_go#decode-instruction

greged93:
Add function to decode a cairo instruction. Resolves #12.


src/primitives/uint.ts
```diff
@@ -29,6 +35,15 @@ export const Uint64ConversionError = {
 export class UnsignedInteger {
   static readonly ZERO: Uint32 = 0 as Uint32;
 
+  // Returns whether the number is a safe unsigned integer,
+  // i.e. a positive number between 0 and 2^32
+  static isUint16(num: number): num is Uint16 {
```
Eikix, line 40: ah I think we should also test if the number is integer in all these utils

src/primitives/uint.ts
```diff
@@ -65,10 +87,49 @@ export class UnsignedInteger {
...
+    }
+    throw Uint64ConversionError;
+  }
+
+  static uint16Rhs(lsh: Uint16, rhs: Uint16): Uint16 {
+    const result = lsh >> rhs;
+    if (this.isUint16(result)) {
+      return result;
+    }
+    throw Uint64ConversionError;
```
Eikix, line 117: why are we throwing here?

src/vm/instruction.ts
```diff
@@ -10,68 +33,303 @@
...
+  pcUpdate: PcUpdate;
+  apUpdate: ApUpdate;
+  fpUpdate: FpUpdate;
+  opcode: Opcode;
+};
+
+export function decodeInstruction(
+  encodedInstruction: Uint64
+): Result<Instruction, VMError> {
+  const highBit = 1n << 63n;
```
Eikix, line 115: I think every single one of these mask, offset, etc. needs to be commented with thorough and pedagogical explanation

src/primitives/uint.ts
```diff
@@ -65,10 +87,49 @@ export class UnsignedInteger {
...
+
+  static uint16And(lhs: Uint16, rhs: Uint16): Uint16 {
+    const result = lhs & rhs;
+    if (this.isUint16(result)) {
+      return result;
+    }
+    throw Uint64ConversionError;
+  }
+
+  static uint16Rhs(lsh: Uint16, rhs: Uint16): Uint16 {
```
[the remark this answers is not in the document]
greged93: oupsi

src/primitives/uint.ts
```diff
@@ -29,6 +35,15 @@ export const Uint64ConversionError = {
...
+  static isUint16(num: number): num is Uint16 {
+    if (num >= 0 && num < 0x10000) {
+      return true;
+    }
+    return false;
+  }
+
   // Returns whether the number is a safe unsigned integer,
   // i.e. a positive number between 0 and 2^32
   static isUint32(num: number): num is Uint32 {
```
Eikix, line 49: same here

src/primitives/uint.ts
```diff
@@ -65,10 +87,49 @@ export class UnsignedInteger {
...
+    }
+    throw Uint64ConversionError;
+  }
+
+  static uint64And(lhs: Uint64, rhs: Uint64): Uint64 {
+    const result = lhs & rhs;
+    if (this.isUint64(result)) {
+      return result;
+    }
+    throw Uint64ConversionError;
```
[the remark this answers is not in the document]
greged93: `anding` two uint64 should always yield a `unint64`, if not something really wrong is happening and we throw

src/vm/instruction.ts
greged93: will add comments for them

src/vm/instruction.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    test('should correctly decode a valid cairo instruction', () => {
+      const shift = 16n;
+      const offDst = 0b010101n;
+      const offOp0 = 0b101010n << shift;
+      const offOp1 = 0b111111n << (2n * shift);
+
+      const flag = 0b0100010000100100n << (3n * shift);
+
+      const encodedInstruction = offDst | offOp0 | offOp1 | flag;
+      const encodedInstructionUint64 =
```
Eikix, line 27: didn't we say it was not so elegant to put the type of a variable in its name?

src/vm/instruction.ts
```diff
@@ -2,6 +2,29 @@
 // Some instructions spread over two words when they use an immediate value, so
 // representing the first one with this struct is enougth.
 
+import { Int16, SignedInteger16 } from 'primitives/int';
+import { Uint64, UnsignedInteger } from 'primitives/uint';
+import { Err, Ok, Result, VMError } from 'result-pattern/result';
+
+export const HighBitSetError: VMError = {
```
Eikix, line 9: Realized we can do this:
```typescript
const DECODING_ERRORS = {
    HIGH_BIT_SET: 'High bit is not zero',
    INVALID_OP1_SRC: 'Invalid Operand 1 Source',
    INVALID_PC_UPDATE: 'Invalid PC Update',
    INVALID_AP_UPDATE: 'Invalid AP Update',
    INVALID_RES_LOGIC: 'Invalid Result Logic',
    INVALID_OPCODE: 'Invalid Opcode'
} as const;
```

src/vm/instruction.test.ts
greged93: I think you wouldn't do it in rust because you can shadow, but can you do the same in ts?

src/vm/instruction.ts
Eikix: And then `return Err({message: `InstructionDecodingError ${DECODING_ERRORS.<whatever>}`)`

src/vm/instruction.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  Instruction,
+  Op1Src,
+  Opcode,
+  PcUpdate,
+  RegisterFlag,
+  ResLogic,
+} from './instruction';
+import { SignedInteger16 } from 'primitives/int';
+
+describe('Instruction', () => {
```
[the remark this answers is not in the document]
greged93: this hasn't been taken from cairo vm go and was build by me

src/vm/instruction.ts
greged93: nice will do

src/vm/instruction.ts
```diff
@@ -10,68 +33,303 @@
...
+    shiftedEncodedInstruction,
+    mask
+  );
+  const offsetOp0Result = SignedInteger16.fromBiased(offsetOp0Uint64);
+  if (offsetOp0Result.isErr()) {
+    return offsetOp0Result;
+  }
+  const offsetOp0 = offsetOp0Result.unwrap();
+
+  var shiftedEncodedInstruction = UnsignedInteger.uint64Rhs(
```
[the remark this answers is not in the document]
Eikix: https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/

src/vm/instruction.ts
```diff
@@ -10,68 +33,303 @@
...
+
+  if ((highBit & encodedInstruction) !== 0n) {
+    return new Err(HighBitSetError);
+  }
+
+  const mask = UnsignedInteger.toUint64(0xffffn).unwrap(); // safe to unwrap
+  const shift = UnsignedInteger.toUint64(16n).unwrap(); // safe to unwrap
+
+  // Get the offset by masking and shifting the encoded instruction
+  const offsetDstUint64 = UnsignedInteger.uint64And(encodedInstruction, mask);
```
Eikix, line 139:
```suggestion
  const offsetDstUint64 = encodedInstruction && mask;
```

Eikix: Since Uint64 extends number, i think we can simply use && and not implement And?

Eikix: Mhm I guess what is nice is that you force both parameters to be Uin64

src/vm/instruction.ts
```diff
@@ -10,68 +33,303 @@
...
+  const offsetDstResult = SignedInteger16.fromBiased(offsetDstUint64);
+  if (offsetDstResult.isErr()) {
+    return offsetDstResult;
+  }
+  const offsetDst = offsetDstResult.unwrap();
+
+  var shiftedEncodedInstruction = UnsignedInteger.uint64Rhs(
```
Eikix, line 146: I'm not sure I like this reassignment with var,

We can:
Define once shiftedEncodedInstruction:
`let shiftedEncodedInstruction = ...`

and then change its value:
`shiftedEncodedInstruction = ...`

Or we can simply redeclare a const every time, that is named after the part of the instruction we're dealing with, i.e.

```typescript
const shiftedEncodedInstructionOp0 = UnsignedInteger.uint64Rhs(encodedInstruction,
    2 * shift
  );
```
```

File: src/primitives/uint.ts (hunk 6 above)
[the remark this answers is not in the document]
Comment (Eikix) [reply src/primitives/uint.ts:125]: ah ok i should never fail, can we add a comment

File: src/vm/instruction.ts (hunk 11 above)
Comment (Eikix) [reply src/vm/instruction.ts:139]: ERRATUM: this is nonsense, please ignore
should be `&` and it's bigint, not number,
and it doesn't work because 
```
const offsetDstUint64 = encodedInstruction & mask
```
is not typed as Uint64

File: src/vm/instruction.test.ts (hunk 9 above)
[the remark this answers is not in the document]
Comment (Eikix) [reply src/vm/instruction.test.ts:16]: nice! 
should we add one test from their codebase as well? for diversity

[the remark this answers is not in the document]
Comment (greged93) [reply src/vm/instruction.test.ts:16]: sure, I can pick a few from them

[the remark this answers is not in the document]
Comment (greged93) [reply src/vm/instruction.test.ts:16]: might check the cairo paper actually, I think they have a couple of cases

File: src/vm/instruction.ts [hunk 13]
```diff
@@ -10,68 +33,303 @@
...
+      break;
+    case 2:
+      apUpdate = ApUpdate.ApUpdateAdd1;
+      break;
+    default:
+      return new Err(InvalidOp1Src);
+  }
+
+  var fpUpdate: FpUpdate;
+  switch (opcode) {
```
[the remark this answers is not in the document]
greged93: this looks nice but isn't the `switch` and `if` `else` easier to understand? If we are aiming for clear and understandable code then maybe this looks a little complicated.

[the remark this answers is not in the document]
Eikix: Okay, it's just that I don't like the `var` declaration before the switch and then assigning it
If we can make the switch / if-else a bit cleaner I'm good

src/vm/instruction.ts
```diff
@@ -10,68 +42,318 @@
...
+      pcUpdate = PcUpdate.PcUpdateJump;
+      break;
+    case 2:
+      pcUpdate = PcUpdate.PcUpdateJumpRel;
+      break;
+    case 4:
+      pcUpdate = PcUpdate.PcUpdateJnz;
+      break;
+    default:
+      return new Err(InvalidOp1Src);
```
Eikix, line 263: wrong error message here and down?

src/vm/instruction.ts
```diff
@@ -248,101 +254,120 @@ export function decodeInstruction(
     UnsignedInteger.uint16And(flags, pcUpdateMask),
     pcUpdateOff
   );
-  var pcUpdate: PcUpdate;
+  let pcUpdate: PcUpdate;
   switch (pcUpdateNum) {
     case 0:
+      // pc = pc + instruction size
```
Eikix, line 260: these are really great!

src/vm/instruction.ts
```diff
@@ -248,101 +254,120 @@ export function decodeInstruction(
...
 
   const resLogicNum = UnsignedInteger.uint16Shr(
     UnsignedInteger.uint16And(flags, resLogicMask),
     resLogicOff
   );
-  var resLogic: ResLogic;
+  let resLogic: ResLogic;
   switch (resLogicNum) {
     case 0:
+      // jnz with res_logic == 0 and pc_update == 3 then unconstrained
```
Eikix, line 286: could you add a bit more information?
I don't understand the use of the word "with" here,
what does unconstrained mean?

what does res = op1 mean? maybe if we use a simple sentence it'll helper the reader

greged93: the issue is that we are only decoding here, not executing yet, but for example `res = op1` I think means `result = operand 1`. Can't be sure, this comes from [cairo paper](https://eprint.iacr.org/2021/1063.pdf#page=33). `with` means if you have `jnz` and res_logic == 0 and pc_update == 3, then the resLogic is unconstrained, meaning you don't care about its value.

Eikix (approved):
lgtm

PR #17, opened 2023-10-20
dev: refactor types

### LINKED ISSUE #16: dev: replace the current types by bigint
We are now using `Uint64`, `Uint32`, `Int16`,... which causes a lot of overhead in order to perform operations on these types, since TS does not allow for operator overload. Replace these types by simple utils which would simply check that the bigint or the number is within the expected range. Each function would check this expected range and return an error if the value is not within range.

greged93:
Refactors the types in order to have more readable function bodies. Resolves #16.


src/vm/instruction.test.ts
```diff
@@ -15,6 +21,54 @@ import { SignedInteger16 } from 'primitives/int';
 
 describe('Instruction', () => {
   describe('decodeInstruction', () => {
+    test('should fail with HighBitSetError', () => {
```
Eikix, line 24: did we try and incorporate tests from cairovmgo, rs, nethermind?

Eikix: (if we do and i think we should take one from each for each test, good to credit as well)

src/vm/instruction.test.ts
```diff
@@ -135,6 +189,29 @@ describe('Instruction', () => {
       expect(instruction).toEqual(expected);
     });
 
+    test('should correctly decode the cairo instruction CALL from LC example', () => {
```
Eikix, line 192:
```suggestion
    test('should correctly decode the cairo instruction CALL - ported from lambdaclass/cairo-vm_in_go', () => {
```

Eikix: that way we can reuse this norm to say
ported from lambdaclass/cairo-vm & nethermind/cairovmgo

src/vm/instruction.test.ts
Eikix: erratum!! i didn't read 5 lines down! 
very nice tests and nice porting of existing tests

src/vm/instruction.ts
```diff
@@ -136,106 +136,75 @@ export function decodeInstruction(
...
-  );
-  const flagsResult = UnsignedInteger.downCastToUint16(
-    shiftedEncodedInstruction
-  );
-  if (flagsResult.isErr()) {
-    return flagsResult;
-  }
-
-  const flags = flagsResult.unwrap();
+  const flags = Number(shiftedEncodedInstruction >> shift);
```
Eikix, line 200: can this Number cast fail?
Why are we casting to a Number here

Eikix: nice naming, helped me understand where we are in the cursor of the whole encoded instruction

Eikix (commented):
nice change! this is way more readable

src/vm/instruction.test.ts
```diff
@@ -20,51 +19,63 @@ import {
 import { SignedInteger16 } from 'primitives/int';
 
 describe('Instruction', () => {
-  describe('decodeInstruction', () => {
+  describe('Instruction.decodeInstruction', () => {
```
Eikix, line 22:
```suggestion
describe('decodeInstruction', () => {
```

greged93: oupsi

Eikix (approved):

PR #22, opened 2023-10-25
dev: option type

greged93:
Implement the option type in order to avoid returning an error when querying an untouched memory cell. Solves #23.


Eikix (approved):
lgtm

PR #25, opened 2023-11-03
feat: compute op1 and deduce op0

greged93:
Updates the error handling and implements the compute op1 and deduce op0.

### DIFF
--- a/src/memory/memoryManager.ts
+++ b/src/memory/memoryManager.ts
@@ -22,19 +22,18 @@ export class MemorySegmentManager {
     address: Relocatable,
     data: MaybeRelocatable[]
   ): Result<Relocatable> {
-    for (let index = 0; index < data.length; index++) {
+    data.forEach((d, index) => {
       const { value: indexUint, error: indexError } =
         UnsignedInteger.toUint32(index);
       if (indexError !== undefined) {
         return { value: undefined, error: indexError };
       }
-
       const { value: nextAddress, error: addrError } = address.add(indexUint);
       if (addrError !== undefined) {
         return { value: undefined, error: addrError };
       }
-      this.memory.insert(nextAddress, data[index]);
-    }
+      this.memory.insert(nextAddress, d);
+    });
 
     const { value: dataLen, error: dataLenError } = UnsignedInteger.toUint32(
       data.length
--- a/src/memory/memoryManager.ts
+++ b/src/memory/memoryManager.ts
@@ -1,6 +1,7 @@
 import { Uint32, UnsignedInteger } from 'primitives/uint';
 import { Memory } from './memory';
 import { MaybeRelocatable, Relocatable } from 'primitives/relocatable';
+import { Result } from 'result/result';
 
 export class MemorySegmentManager {
   private segmentSizes: Map<Uint32, Uint32>;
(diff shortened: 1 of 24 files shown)


src/result/error.ts
```diff
@@ -0,0 +1,28 @@
...
+  VMError: 'VMError',
+  InstructionError: 'InstructionError',
+  RunContextError: 'RunContextError',
+} as const;
+
+export class BaseError extends Error {
+  type: string;
+  message: string;
+
+  constructor(type: string, message: string) {
```
Eikix, line 16:
```suggestion
  constructor(type: typeof ErrorType, message: string) {
```

src/result/error.ts
```diff
@@ -0,0 +1,28 @@
...
+  Int16Error: 'Int16Error',
+  UintError: 'UintError',
+  MemoryError: 'MemoryError',
+  VMError: 'VMError',
+  InstructionError: 'InstructionError',
+  RunContextError: 'RunContextError',
+} as const;
+
+export class BaseError extends Error {
+  type: string;
```
Eikix, line 13:
```suggestion
  type: typeof ErrorType;
```

src/vm/virtualMachine.ts
```diff
@@ -28,20 +25,113 @@ export class VirtualMachine {
...
+
     // return this.runInstruction();
     throw new Error('TODO: Not Implemented');
   }
 
-  computeOperands(instruction: Instruction) {}
+  // Compute the operands of an instruction. The VM can either
+  // fetch them from memory and if it fails to do so, it can
+  // deduce them from the instruction itself.
+  computeOperands(instruction: Instruction): Result<void> {
```
Eikix, line 52: why does it return void?
is it a todo?

src/memory/memoryManager.ts
```diff
@@ -17,18 +18,38 @@ export class MemorySegmentManager {
     return ptr;
   }
 
-  loadData(address: Relocatable, data: MaybeRelocatable[]): Relocatable {
+  loadData(
+    address: Relocatable,
+    data: MaybeRelocatable[]
+  ): Result<Relocatable> {
     for (let index = 0; index < data.length; index++) {
```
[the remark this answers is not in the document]
greged93: mmmmm depends, isn't a `for` loop a little more readable?

src/vm/virtualMachine.ts
greged93: yes, currently working on the rest of the computation that will return the operands

src/memory/memoryManager.ts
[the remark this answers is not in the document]
Eikix: i feel as though forEach is slightly more readable imo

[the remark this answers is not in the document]
greged93: will switch to for then

src/result/error.ts
greged93: I had to do this `type ErrorTypeKey = (typeof ErrorType)[keyof typeof ErrorType];` becase you are actually passing the value of the ErrorType object.

src/result/memory.ts
```diff
@@ -0,0 +1,8 @@
+export const UnknownAddressError =
+  'Access memory at unknown or uninitialized address';
+export const WriteOnceError =
+  'Write existing memory. Can only write to memory once.';
+export const SegmentIncrementError = 'Error incrementing number of segments';
+export const InstructionError = 'Instruction must be a Field Element';
```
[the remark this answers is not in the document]
greged93: this is actually used in the virtual machine: we get the next instruction which can be found in the pc memory segment at the current pc offset. This returns a maybe relocatable. If this maybe relocatable is not a felt, we throw this error

Eikix (approved):

PR #26, opened 2023-11-10
feat: compute operands suite

greged93:
Add additional helpers functions for the compute operands + complete compute operands.

### DIFF
--- a/src/vm/virtualMachine.ts
+++ b/src/vm/virtualMachine.ts
@@ -8,8 +8,11 @@ import { BaseError, ErrorType } from 'result/error';
 import { EndOfInstructionsError, InstructionError } from 'result/memory';
 import { Result } from 'result/result';
 
-export const DeductionError = {
-  message: 'VMError: failed to deduce operands',
+export type Operands = {
+  op0: MaybeRelocatable | undefined;
+  op1: MaybeRelocatable | undefined;
+  res: MaybeRelocatable | undefined;
+  dst: MaybeRelocatable | undefined;
 };
 
 export class VirtualMachine {
@@ -53,7 +56,8 @@ export class VirtualMachine {
   // Compute the operands of an instruction. The VM can either
   // fetch them from memory and if it fails to do so, it can
   // deduce them from the instruction itself.
-  computeOperands(instruction: Instruction): Result<void> {
+  computeOperands(instruction: Instruction): Result<Operands> {
+    let res: MaybeRelocatable | undefined;
     // Compute the destination address based on the dstReg and
     // the offset
     const { value: dstAddr, error: dstError } = this.runContext.computeAddress(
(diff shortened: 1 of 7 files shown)


src/vm/virtualMachine.ts
```diff
@@ -83,9 +91,83 @@ export class VirtualMachine {
...
+    // If dst is undefined, then we can deduce it from the instruction and res
+    if (dst === undefined) {
+      const { value: deducedDst, error: deducedDstError } = this.deduceDst(
+        instruction,
+        res
+      );
+      if (deducedDstError !== undefined) {
+        return { value: undefined, error: deducedDstError };
+      }
+      if (deducedDst !== undefined) {
```
Eikix, line 156: can deducedDst be undefined here? we checked for errors, it feels weird to check again

src/vm/virtualMachine.ts
```diff
@@ -83,9 +91,83 @@ export class VirtualMachine {
...
+        this.segments.memory.insert(dstAddr, deducedDst);
+      }
+      dst = deducedDst;
```
Eikix, line 159: especially because if it is undefined, we don't insert, but we still overwrite dst at line 159

src/vm/virtualMachine.ts
```diff
@@ -59,7 +67,7 @@ export class VirtualMachine {
     if (dstError !== undefined) {
       return { value: undefined, error: dstError };
     }
-    const dst = this.segments.memory.get(dstAddr);
+    let dst = this.segments.memory.get(dstAddr);
```
Eikix, line 70: we need to document in which cases dst is already defined in memory, as opposed to it being undefined, where we'd need to deduce it

src/vm/virtualMachine.ts
greged93: but deduceDst can return undefined error and value, you have no error but it couldn't deduce the value

src/vm/virtualMachine.ts
```diff
@@ -83,9 +91,83 @@ export class VirtualMachine {
...
+        return { value: undefined, error: deducedOp0Error };
+      }
+      const [deducedOp0, deducedRes] = deducedValues;
+      if (deducedOp0 !== undefined) {
+        this.segments.memory.insert(op0Addr, deducedOp0);
+      }
+      res = deducedRes;
+    }
+
+    // If op1 is undefined, then we can deduce it from the instruction, dst and op0
```
Eikix, line 114: can we document in which cases op1 is undefined
also in comment I think we should use the full name of variables

Eikix (commented):
Nice PR! globally in agreement with the impl, emitted a "wishlist" on documenting the codebase and ability to ELI5

src/vm/virtualMachine.ts
Eikix: why?

greged93: Well there is a `if` branch on `dst === undefined` so in this current branch, dst in already undefined. We are trying to deduce it but only want to write to memory if the deduction worked.

src/vm/virtualMachine.ts
greged93: Honestly, I have no idea in which cases op1 can be undefined :/ I guess it could just be if the memory cell you read is empty

Eikix (approved):
lgtm

PR #30, opened 2023-11-16
feat: cairo runner

### LINKED ISSUE #18: dev: implement `runInstruction` for the `vm`
```[tasklist]
### Tasks
- [ ] https://github.com/kkrt-labs/cairo-vm-ts/issues/19
- [ ] https://github.com/kkrt-labs/cairo-vm-ts/issues/20
- [ ] https://github.com/kkrt-labs/cairo-vm-ts/issues/21
```

greged93:
Adds the cairo runner to the code base, and allows to run fibonnaci! Resolves #18


src/memory/memoryManager.ts
```diff
@@ -22,7 +24,7 @@ export class MemorySegmentManager {
     address: Relocatable,
     data: MaybeRelocatable[]
   ): Result<Relocatable> {
-    data.forEach((d, index) => {
```
greged93, line 25: Removed the forEach, I know you like it @Eikix but it was actually preventing errors to be propagated.

src/memory/memoryManager.ts
```diff
@@ -54,6 +56,33 @@ export class MemorySegmentManager {
     return address.add(dataLen);
   }
 
+  // Insert a value in the memory at the given address and increase
+  // the segment size by 1.
+  insert(address: Relocatable, value: MaybeRelocatable): void {
```
greged93, line 61: Moved insert here and completely removed `insert` from Memory. This is done in order to be able to keep track of the segment length, which wasn't possible when inserting data via the Memory.

src/primitives/relocatable.ts
```diff
@@ -31,24 +31,27 @@ export class Relocatable {
...
+  add(other: Uint32): Ok<Relocatable>;
   add(other: MaybeRelocatable): Result<MaybeRelocatable>;
 
   add(other: MaybeRelocatable | Uint32): Result<MaybeRelocatable> {
     if (other instanceof Felt) {
-      const { value: num, error } = other.toUint32();
-      if (error !== undefined) {
-        return { value: undefined, error };
+      const offsetFelt = new Felt(BigInt(this.getOffset()));
+      const { value: sum, error: sumError } = offsetFelt.add(other);
```
greged93, line 41: First add on the Felts, in order to wrap around. Then convert to uint32

src/vm/program.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export type Identifier = z.infer<typeof Identifier>;
+
+const Program = z.object({
+  attributes: z.any(),
+  builtins: z.array(z.string()),
+  compiler_version: z.string(),
+  data: z
+    .array(z.string())
+    .transform((value) => value.map((v) => new Felt(BigInt(v)))),
```
Eikix, line 44: do we need this map?

Makefile
```diff
@@ -0,0 +1,8 @@
+CAIRO_0_PATH=cairo_programs/cairo_0
```
Eikix, line 1: do we have a strategy for cairo1 files as well?
with another cairo-compile binary? or rather scarb?
it can wait

.trunk/trunk.yaml
ClementWalter, line 1: can we add the cairo-format formatter as well (you can copy/paste from kakarot)

cairo_programs/cairo_0/fibonacci.cairo
ClementWalter, line 1: is this the fibonacci implementation used in other repo? as it's not the most efficient one

ClementWalter: see also the cairo paper, section 3.4 Cairo programs

src/runners/cairoRunner.ts
```diff
@@ -0,0 +1,175 @@
...
+    this.initialFp = new Relocatable(0, 0);
+    this.finalPc = new Relocatable(0, 0);
+    this.mainOffset = mainOffset;
+  }
+
+  // Initialize the runner with the program and stack.
+  initialize(): Result<Relocatable> {
+    this.initializeSegments();
+    const result = this.initializeMainEntrypoint();
+    this.initializeVm();
```
ClementWalter, line 46: this is confusing because we get as intermediary result `result` which is eventually returned. Why not 

```
    this.initializeSegments();
    this.initializeVm();
    return this.initializeMainEntrypoint();    
```

but I think that the content of these functions should be inlined here, they are small and it doesn't help readability

ClementWalter: after pulling the PR locally, I definitely think that you should inline everything and remove all the `initializeX` functions

src/vm/virtualMachine.ts
ClementWalter, line 1: you case use `default` in all the switch instead of leaving the switch to `return`

ClementWalter: also plz put the early return in the branches to keep the main case at the most top level

ClementWalter (commented):
the `src/vm/virtualMachine.ts` file is hard to read

src/runners/cairoRunner.ts
greged93: Mmmmm that might be a lot of code inlined into one function, worried this will not improve the lisibility nor the testing.

cairo_programs/cairo_0/fibonacci.cairo
greged93: just tried with low level, it's nice, can update if you want

ClementWalter (approved):

ClementWalter (approved):

PR #33, opened 2023-11-17
Refactor: remove Result pattern

Eikix:
We throw everywhere instead of handling errors


Eikix:
Next PRs:
1.  Will be to try and simply number conversions to further reduce number of lines
2. We might want to reduce "safety" by removing verbose getters and making Class fields public

Eikix:
> Next PRs:
> 
> 1. Will be to try and simply number conversions to further reduce number of lines
> 2. We might want to reduce "safety" by removing verbose getters and making Class fields public

Added it in this PR!

src/memory/memory.ts
```diff
@@ -17,16 +17,15 @@ export class Memory {
...
-    const { value: numSegments, error } = UnsignedInteger.toUint32(
-      this.numSegments + 1
-    );
-    if (error !== undefined) {
+    try {
+      this.numSegments += 1;
+      UnsignedInteger.ensureUint32(this.numSegments);
+    } catch (err) {
       throw new MemoryError(SegmentIncrementError);
     }
```
ClementWalter, line 25: why do we enforce numSegment to be uint32?

src/memory/memoryManager.test.ts
```diff
@@ -26,39 +26,33 @@ describe('MemoryManager', () => {
...
     });
   });
-  describe('loadData', () => {
+  describe('setData', () => {
     test('should return the final state of the pointer', () => {
       const memoryManager = new MemorySegmentManager();
       memoryManager.addSegment();
       const address = new Relocatable(0, 0);
-      const { value: loadedAddress } = memoryManager.loadData(address, DATA);
+      const newAddress = memoryManager.setData(address, DATA);
```
ClementWalter, line 37: I know that even though I don't like it, js does return the latest value when doing `arr.push(1, 2, 3)` (ie.

```
> let a = []
undefined
> a.push(1, 2)
2
> a
[ 1, 2 ]
```

but here it's returning the next free memory pointer, do we really want this? When I read something like `this.mutateWith(args0, args1)` I don't expect a returned value

src/memory/memoryManager.ts
```diff
@@ -20,54 +19,28 @@ export class MemorySegmentManager {
...
     this.insert_inner(address, value);
     if (address.getOffset() > this.getSegmentSize(address.getSegmentIndex())) {
-      const { value: newSize, error } = UnsignedInteger.toUint32(
-        address.getOffset() + UnsignedInteger.ONE_UINT32
-      );
-      if (error !== undefined) {
-        throw error;
-      }
-      this.segmentSizes.set(address.getSegmentIndex(), newSize);
+      UnsignedInteger.ensureUint32(address.getOffset() + 1);
```
ClementWalter, line 42: do we really need this? a segment could be just an array and in js, it's possible to do

```
a[0] = 1
a[10] = 1
```

and `a.length` would return the `segmentSize`

src/memory/memoryManager.ts
```diff
@@ -1,12 +1,11 @@
...
 import { MaybeRelocatable, Relocatable } from 'primitives/relocatable';
-import { Result } from 'result/result';
-import { MemoryError, WriteOnceError } from 'result/memory';
-import { SegmentError } from 'result/primitives';
+import { MemoryError, WriteOnceError } from 'errors/memory';
+import { SegmentError } from 'errors/primitives';
 
 export class MemorySegmentManager {
-  private segmentSizes: Map<Uint32, Uint32>;
+  private segmentSizes: Map<number, number>;
```
ClementWalter, line 8: so I think that this is useless if we use a list of list in the memory

src/primitives/relocatable.ts
```diff
@@ -1,147 +1,77 @@
...
-  add(other: Felt): Result<Relocatable>;
-  add(other: Uint32): Ok<Relocatable>;
-  add(other: MaybeRelocatable): Result<MaybeRelocatable>;
-
-  add(other: MaybeRelocatable | Uint32): Result<MaybeRelocatable> {
+  add(other: Felt): Relocatable;
+  add(other: number): Relocatable;
+  add(other: Relocatable): never;
+  add(other: MaybeRelocatable): MaybeRelocatable;
+  add(other: MaybeRelocatable | number): MaybeRelocatable {
```
ClementWalter, line 27: maybe a ts question, but it seeems like the last signature is redundant?

src/run-context/runContext.ts
```diff
@@ -30,18 +29,13 @@ export class RunContext {
...
 
-  incrementPc(instructionSize: Uint32): Result<void> {
-    const { value: newPc, error: newPcError } = this.pc.add(instructionSize);
-    if (newPcError !== undefined) {
-      return { value: undefined, error: newPcError };
-    }
-    this.pc = newPc;
-    return { value: undefined, error: undefined };
+  incrementPc(instructionSize: number): void {
+    this.pc = this.pc.add(instructionSize);
```
ClementWalter, line 33: should the `pc` be just a number so that we could write `this.pc + size`?

ClementWalter (approved):

src/memory/memoryManager.ts
greged93: Is there an advantage to use list of list? I think the cairo vm in zig was doing list of list, the issue is that you end up reserving real memory for any memory holes, whereas map will just skip the holes.
Example:
[[empty, empty, empty, Relocatable(1,1)]] vs {"0:3": Relocatable(1,1)}

src/primitives/relocatable.ts
```diff
@@ -1,147 +1,77 @@
...
-        error: new PrimitiveError(ForbiddenOperation),
-      };
+      throw new PrimitiveError(ForbiddenOperation);
     }
 
-    return {
-      value: new Relocatable(this.getSegmentIndex(), this.getOffset() + other),
-      error: undefined,
-    };
+    UnsignedInteger.ensureUint32(other);
```
greged93, line 39: Think you can remove this since you check it in the constructor

src/primitives/relocatable.ts
```diff
@@ -1,147 +1,77 @@
...
-      error: new PrimitiveError(ForbiddenOperation),
-    };
-  }
-
-  div(_: MaybeRelocatable | Uint32): Err {
-    return {
-      value: undefined,
-      error: new PrimitiveError(ForbiddenOperation),
-    };
+    UnsignedInteger.ensureUint32(other);
```
greged93, line 73: same here, you can remove the uint32 check

src/run-context/runContext.ts
```diff
@@ -55,9 +49,10 @@ export class RunContext {
   // apply to the source and the operand 0.
   computeOp1Address(
     op1Src: Op1Src,
-    op1Offset: Int16,
+    op1Offset: number,
     op0: MaybeRelocatable | undefined
-  ): Result<Relocatable> {
+  ): Relocatable {
+    SignedInteger16.ensureInt16(op1Offset);
```
greged93, line 55: I don't know if removing all the primitives type is a good idea. Here refactoring was easy because you know from before (because it was typed before) that you needed to add this `ensureInt16`. Could we keep the `Int16`, but have it be just a number? So just use type aliasing? Would make it more clear also for the reader and the dev imo

greged93 (commented):
I think I agree with most of the refactoring, but not really with the complete removal of the `primitives` types. I would keep `Uint32`, `Int16` but change them to full type aliasing. it makes it imo clearer to the reader and also for us, instead of having to scan the code to see if it contains `ensure....`

src/memory/memory.ts
Eikix: I think because relocatable is numSegment: uint32, offset: u32

src/primitives/relocatable.ts
Eikix: > maybe a ts question, but it seeems like the last signature is redundant?

the last signature is the implementation, it  must be compatible with all of the above, so its a big union

src/memory/memoryManager.ts
Eikix: actually i think this makes sense, in the effort of making everything as simple as possible

Eikix:
creating an issue to track the next changes

PR #36, opened 2023-11-24
Feat/spring cleaning

Eikix:
- make memory list of list
- delete memory manager
- transform Uint32 into Uint53 -> use Number.MAX_SAFE_INTEGER as a usize high cap

challengeable:
- bug in fibonacci test: it appears executionEnd is last known AP -> whereas AP points to the latest free cell, so technically, shouldn't last fibonacci be AP.sub(1) -> since it's the latest "taken" memory cell? see per-ultimate commit


src/runners/cairoRunner.test.ts
```diff
@@ -25,11 +24,15 @@ describe('cairoRunner', () => {
...
-      runner.runUntilPc(finalPc);
-      const executionSize = runner.getVm().segments.getSegmentSize(1 as Uint32);
+      runner.runUntilPc(finalPc, true);
+      const executionSize = runner.getVm().memory.getSegmentSize(1);
       const executionEnd = runner.getExecutionBase().add(executionSize);
 
-      expect(runner.getVm().segments.memory.get(executionEnd)).toEqual(
+      console.log(
+        `Execution end: ${executionEnd.getSegmentIndex()}:${executionEnd.getOffset()}`
+      );
```
greged93, line 33: you left a log here

src/runners/cairoRunner.ts
```diff
@@ -43,52 +42,66 @@ export class CairoRunner {
...
+  runUntilPc(finalPc: Relocatable, verbose: boolean = false): void {
+    while (this.vm.runContext.pc.getOffset() < finalPc.getOffset()) {
       this.vm.step();
+      if (verbose) {
+        console.log(
+          `AP: ${this.vm.runContext.ap.getSegmentIndex()}:${this.vm.runContext.ap.getOffset()}`
+        );
+        console.log(
+          `[ap - 1]: ${this.vm.memory.get(this.vm.runContext.ap.sub(1))}`
+        );
```
greged93, line 54: Maybe print Ap, Fp, Pc and then [ap -1]? should we also print [fp]?

greged93 (approved):
Lgtm

PR #46, opened 2024-04-24
refactor: errors

zmalatrax:
Move error handling from a few classes with specific strings to classes-only.
It allow for a more granular error management.

Currently unused errors have been deleted.

Too verbose errors were renamed:
- `InvalidOperandOneSource` -> `InvalidOp1Source`
- `InvalidOperand0` -> `InvalidOp0`
- `InvalidOperand1` -> `InvalidOp1`

### DIFF
--- a/src/errors/memory.ts
+++ b/src/errors/memory.ts
@@ -1,14 +1,20 @@
-import { MaybeRelocatable } from 'primitives/relocatable';
+import { MaybeRelocatable, Relocatable } from 'primitives/relocatable';
 
 export class MemoryError extends Error {}
 
-/** Write existing memory. Can only write to memory once */
+/** Read a different value (`newValue`) than the already constrained value (`oldValue`) at `address` */
 export class InconsistentMemory extends MemoryError {
+  public readonly address: Relocatable;
   public readonly oldValue: MaybeRelocatable;
   public readonly newValue: MaybeRelocatable;
 
-  constructor(oldValue: MaybeRelocatable, newValue: MaybeRelocatable) {
+  constructor(
+    address: Relocatable,
+    oldValue: MaybeRelocatable,
+    newValue: MaybeRelocatable
+  ) {
     super();
+    this.address = address;
     this.oldValue = oldValue;
     this.newValue = newValue;
   }
--- a/src/errors/memory.ts
+++ b/src/errors/memory.ts
@@ -1,9 +1,5 @@
 export class MemoryError extends Error {}
 
-export const UnknownAddressError =
-  'Access memory at unknown or uninitialized address';
 export const WriteOnceError =
   'Write existing memory. Can only write to memory once.';
-export const SegmentIncrementError = 'Error incrementing number of segments';
 export const InstructionError = 'Instruction must be a Field Element';
-export const DeductionError = 'Failed to deduce operands';
--- a/src/errors/memory.ts
+++ b/src/errors/memory.ts
@@ -3,17 +3,16 @@ export class MemoryError extends Error {}
 /** Write existing memory. Can only write to memory once */
 export class WriteOnceError extends MemoryError {}
 
-/** Trying to write on the segment `segment_index`
- * while only `segment_number` are currently defined
+/** Trying to read on a segment that is not accessible (`segmentIndex >= segmentNumber`)
  */
-export class WriteInvalidSegment extends MemoryError {
-  public readonly segment_index: number;
-  public readonly segment_number: number;
+export class SegmentOutOfBounds extends MemoryError {
+  public readonly segmentIndex: number;
+  public readonly segmentNumber: number;
 
   constructor(segment_index: number, segment_number: number) {
     super();
-    this.segment_index = segment_index;
-    this.segment_number = segment_number;
+    this.segmentIndex = segment_index;
+    this.segmentNumber = segment_number;
   }
 }
 
--- a/src/errors/memory.ts
+++ b/src/errors/memory.ts
@@ -1,5 +1,21 @@
 export class MemoryError extends Error {}
 
-export const WriteOnceError =
-  'Write existing memory. Can only write to memory once.';
-export const InstructionError = 'Instruction must be a Field Element';
+/** Write existing memory. Can only write to memory once */
+export class WriteOnceError extends MemoryError {}
+
+/** Trying to write on the segment `segment_index`
+ * while only `segment_number` are currently defined
+ */
+export class WriteInvalidSegment extends MemoryError {
+  public readonly segment_index: number;
+  public readonly segment_number: number;
+
+  constructor(segment_index: number, segment_number: number) {
+    super();
+    this.segment_index = segment_index;
+    this.segment_number = segment_number;
+  }
+}
+
+/** Instruction must be a Field Element */
+export class InstructionError extends MemoryError {}
--- a/src/errors/memory.ts
+++ b/src/errors/memory.ts
@@ -1,10 +1,20 @@
+import { MaybeRelocatable } from 'primitives/relocatable';
+
 export class MemoryError extends Error {}
 
 /** Write existing memory. Can only write to memory once */
-export class WriteOnceError extends MemoryError {}
+export class InconsistentMemory extends MemoryError {
+  public readonly oldValue: MaybeRelocatable;
+  public readonly newValue: MaybeRelocatable;
 
-/** Trying to read on a segment that is not accessible (`segmentIndex >= segmentNumber`)
- */
+  constructor(oldValue: MaybeRelocatable, newValue: MaybeRelocatable) {
+    super();
+    this.oldValue = oldValue;
+    this.newValue = newValue;
+  }
+}
+
+/** Trying to read on a segment that is not accessible (`segmentIndex >= segmentNumber`) */
 export class SegmentOutOfBounds extends MemoryError {
   public readonly segmentIndex: number;
   public readonly segmentNumber: number;
--- a/src/memory/memory.test.ts
+++ b/src/memory/memory.test.ts
@@ -60,14 +60,14 @@ describe('Memory', () => {
       const memory = new Memory();
       memory.addSegment();
       const address = new Relocatable(0, 10);
-      memory.read(address, DATA[0]);
+      memory.assertEq(address, DATA[0]);
 
       expect(memory.get(address)).toEqual(DATA[0]);
     });
     test('should throw SegmentOutOfBounds on access to uninitialized segment', () => {
       const memory = new Memory();
       const address = new Relocatable(1, 10);
-      expect(() => memory.read(address, DATA[0])).toThrow(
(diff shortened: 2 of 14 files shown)


src/errors/memory.ts
```diff
@@ -1,9 +1,21 @@
...
 
-export const UnknownAddressError =
-  'Access memory at unknown or uninitialized address';
-export const WriteOnceError =
-  'Write existing memory. Can only write to memory once.';
-export const SegmentIncrementError = 'Error incrementing number of segments';
-export const InstructionError = 'Instruction must be a Field Element';
-export const DeductionError = 'Failed to deduce operands';
+/** Write existing memory. Can only write to memory once */
+export class WriteOnceError extends MemoryError {}
```
ClementWalter, line 4: See the Cairo paper

> It turns out that, for most of the memory accesses in programs, a Nondeterministic Read-Only Memory is sufficient (for example, Section 6 explains
how a function stack may be implemented in the read-only model), and in
places where it does not suffice, it’s possible to simulate a full read-write
memory using it (see Section 8.5). Therefore, Cairo uses a nondeterministic
read-only memory as its memory model.

src/errors/memory.ts
```diff
@@ -1,9 +1,21 @@
...
+
+/** Trying to write on the segment `segment_index`
+ * while only `segment_number` are currently defined
+ */
+export class WriteInvalidSegment extends MemoryError {
+  public readonly segment_index: number;
```
ClementWalter, line 10: typescript case is camelCase, you should add trunk in the CI

src/errors/memory.ts
```diff
@@ -1,9 +1,21 @@
...
-export const WriteOnceError =
-  'Write existing memory. Can only write to memory once.';
-export const SegmentIncrementError = 'Error incrementing number of segments';
-export const InstructionError = 'Instruction must be a Field Element';
-export const DeductionError = 'Failed to deduce operands';
+/** Write existing memory. Can only write to memory once */
+export class WriteOnceError extends MemoryError {}
+
+/** Trying to write on the segment `segment_index`
+ * while only `segment_number` are currently defined
```
ClementWalter, line 7: so it's much like a segments out of bounds?

src/memory/memory.test.ts
```diff
@@ -68,15 +67,19 @@ describe('Memory', () => {
     test('should throw SegmentError on uninitialized segment', () => {
       const memory = new Memory();
       const address = new Relocatable(1, 10);
-      expect(() => memory.write(address, DATA[0])).toThrow(SegmentError);
+      expect(() => memory.write(address, DATA[0])).toThrow(
+        new WriteInvalidSegment(address.segment, memory.getSegmentNumber())
```
ClementWalter, line 71: so the error is more "UninitializedSegment" ?

src/errors/memory.ts
ClementWalter: to be more accurate, the prover can write once, but the cairo vm is read-only if this makes sense. So the error may stay `WriteOnce` depending on the context

src/primitives/felt.test.ts
```diff
@@ -1,20 +1,15 @@
 import { test, expect, describe } from 'bun:test';
 import { Felt } from './felt';
-import { OutOfRangeBigInt, PrimitiveError } from 'errors/primitives';
-import { Relocatable } from './relocatable';
+import { OutOfRangeBigInt } from 'errors/primitives';
 
 describe('Felt', () => {
   describe('constructor', () => {
     test('should throw if initializing a felt with a negative inner', () => {
```
ClementWalter, line 7: do we really want to throw in this case? isn't just wrapping to `PRIME - n`?

src/errors/memory.ts
zmalatrax: Indeed, `WriteOnce` doesn't fit perfectly

Renaming it `InconsistentMemoryError` as in the Rust/Python VM seems better as it conveys the read-only status of the memory in the cairo vm:
- Once a cell value is constrained (i.e. not undefined), attempting to make read another value from this cell throws `InconsistentMemoryError`

But I would also rename the [`write()`](https://github.com/kkrt-labs/cairo-vm-ts/blob/49159f2cc352c06fda4e4c5cb306ff3228150961/src/memory/memory.ts#L33) function then

Is `read()` more meaningful? (python vm defines `__setitem__` while rust vm declares `insert()` which is equivalent to `write()` imo

src/primitives/felt.test.ts
zmalatrax: It is better to wrap around PRIME at constructor level.
Plus it allows simplifying the different operator methods (add, sub, mul)

src/memory/memory.test.ts
```diff
@@ -56,27 +55,40 @@ describe('Memory', () => {
     });
   });
 
-  describe('write', () => {
-    test('should write a value in the memory at the given address', () => {
+  describe('read', () => {
```
ClementWalter, line 58: `describe('assertEq', () => {`

and then 

```
test('should write when the value is undefined', () => {
test('should not throw when the given value is the one stored', () => {
test('should throw when the given value is not the one stored', () => {
test('should throw when the segment is not initialized', () => {
```

src/memory/memory.ts
```diff
@@ -24,24 +23,32 @@ export class Memory {
 
   setData(address: Relocatable, data: MaybeRelocatable[]): void {
```
ClementWalter, line 24: it looks like we alternate with `data` and `value`, if we go with `value` or `segmentValue`, then let's use `value` and `values` everywhere

src/primitives/felt.test.ts
```diff
@@ -1,20 +1,24 @@
...
-import { Relocatable } from './relocatable';
 
 describe('Felt', () => {
   describe('constructor', () => {
-    test('should throw if initializing a felt with a negative inner', () => {
-      expect(() => new Felt(-10n)).toThrow(
-        new PrimitiveError(OutOfRangeBigInt)
+    test('should properly initialize a Felt with a negative bigint as parameter', () => {
+      const negativeUnderPrime = -(Felt.PRIME + 10n);
+      const negativeEvenUnderPrime = -(Felt.PRIME + Felt.PRIME + 15n);
```
ClementWalter, line 8: I don't understand these choices. Why not just `-10` or `-10n`?
+ never do two tests in the same test
+ even is "pair" while you use `-15`

src/primitives/felt.test.ts
```diff
@@ -1,20 +1,24 @@
...
 import { Felt } from './felt';
-import { OutOfRangeBigInt, PrimitiveError } from 'errors/primitives';
-import { Relocatable } from './relocatable';
 
 describe('Felt', () => {
   describe('constructor', () => {
-    test('should throw if initializing a felt with a negative inner', () => {
-      expect(() => new Felt(-10n)).toThrow(
-        new PrimitiveError(OutOfRangeBigInt)
+    test('should properly initialize a Felt with a negative bigint as parameter', () => {
```
ClementWalter, line 6: it looks like you want to "parametrize"  your test to have several cases covered, see "jest parameterized test" in google

src/primitives/felt.ts
```diff
@@ -1,61 +1,54 @@
...
   private inner: bigint;
   static PRIME: bigint =
     0x800000000000011000000000000000000000000000000000000000000000001n;
   static ZERO: Felt = new Felt(0n);
   constructor(_inner: bigint) {
-    if (_inner < 0n || _inner > Felt.PRIME) {
-      throw new PrimitiveError(OutOfRangeBigInt);
-    }
-    this.inner = _inner;
+    while (_inner < 0n) _inner += Felt.PRIME;
```
ClementWalter, line 11: why not just `this.inner = _inner % Felt.PRIME` ?

src/primitives/felt.ts
```diff
@@ -1,61 +1,54 @@
...
 
   div(other: MaybeRelocatable): Felt {
     if (!(other instanceof Felt) || other.inner === 0n) {
-      throw new PrimitiveError(ForbiddenOperation);
+      throw new ForbiddenOperation();
     }
-    let result = this.inner / other.inner;
-    return new Felt(result);
+
+    return new Felt(this.inner / other.inner);
```
ClementWalter, line 44: this is not the proper felt division, which should return x such that `other * x = inner`

src/primitives/felt.ts
zmalatrax: The `inner` attribute should be positive-only, but Javascript modulo returns the sign of the dividend
(`-9 % 5 = 1`) (wanted behaviour)
(`-9 % 5 = -4`) (JavaScript behaviour)

To have postive-only modulo with `%` there is this [trick](https://web.archive.org/web/20090717035140if_/javascript.about.com/od/problemsolving/a/modulobug.htm):
`this.inner = (_inner % Felt.PRIME) + Felt.PRIME) % Felt.PRIME)`

But I feel like it reduces readability

One suggestion could be

```suggestion
    if (_inner < 0n) {
      _inner = _inner % Felt.PRIME + Felt.PRIME
    }
```

src/primitives/felt.test.ts
zmalatrax: I wanted to check that for `n = -(q*Felt.PRIME + r)`, with `q,r > 0`, the wrapping works as expected, but it isn't that meaningful. I'll stick to `-10n`

`EvenUnder` is bad wording, it wasn't about the parity

src/primitives/felt.ts
zmalatrax: Indeed, it is not a proper felt division, which should be the multiplicative inverse:
$a / b = a * b^{-1} = x$

```suggestion
const inv_other = other.inv()
return this.inner.mul(inv_other)
```

With `inv()` being a method to compute the inverse of a felt. The easiest & widely known algorithm to implement is the extended euclidean algorithm

src/memory/memory.ts
zmalatrax: `value` & `values` make more sense imo, but we'll still have an occurence of `data` in the project, as it is a field name of Cairo Zero compiled json.

No big deal though, as it can be limited to the (Cairo Zero) Program type and we use `value`/`values` everywhere else 🤔

src/primitives/felt.test.ts
```diff
@@ -1,21 +1,30 @@
...
+      }
+    );
+
+    test.each([
+      [Felt.PRIME, new Felt(0n)],
+      [Felt.PRIME + 10n, new Felt(10n)],
+      [Felt.PRIME + 15n, new Felt(15n)],
+      [Felt.PRIME * 2n + 10n, new Felt(10n)],
+    ])(
+      'should initialize a Felt with r = x % PRIME for x a bigint equal to q*PRIME + r',
```
ClementWalter, line 23: you can merge test cases

ClementWalter (approved):

ClementWalter (approved):

PR #47, opened 2024-04-24
dev: remove custom integer classes

zmalatrax:
The use of custom classes to enforce signed and unsigned integers of a certain size reduces the readability of the code.
Plus it doesn't make much sense in TypeScript, it feels like trying to add C/Rust fixed-size integers to Typescript

- Converting a Felt to `number` and `bigint` must be doable:
  - `Number(felt)` works as expected because `toString()` is overridden (see [javascript object coercition](https://2ality.com/2012/11/coercing-objects.html))
  - `BigInt(value: string | number | bigint | boolean):  bigint` has a fixed signature, we must define a method to return the bigint value from Felt, which is done by `Felt.toBigInt(): bigint`

- Encoded instructions are defined on 64 bits. We can check that the instruction is indeed a number < $2^{64} - 1$, [as in the python VM ](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/lang/compiler/instruction.py#L121-L123), or simply ignore the bits > 64. The encoded instruction is decoded with masks, solely working on the 64 LSBs.

- Values from the encoded instructions are truncated with masks of the expected size, there is no need to check that the offsets etc. are of the right size

- `fromBiased` should be a method of Instruction


src/primitives/felt.test.ts
```diff
@@ -21,7 +21,7 @@ describe('Felt', () => {
   describe('conversions', () => {
     test('should convert correctly a felt to a number if inner is below Javascript max safe integer', () => {
```
ClementWalter, line 22: update the description

ClementWalter (approved):

PR #48, opened 2024-04-25
Some improvements on primitives

tcoratger:
- Some functions are a bit refactored in primitive components.
- The `isFelt` and `isRelocatable` utils are used whenever possible.
- These two utils are integrated into a namespace around `MaybeRelocatable`, which seems more logical to me because we want to know if a `MaybeRelocatable` is a Felt or a Relocatable.


src/primitives/felt.ts
```diff
@@ -19,14 +19,14 @@ export class Felt {
   }
 
   add(other: MaybeRelocatable): Felt {
-    if (!(other instanceof Felt)) {
-      throw new PrimitiveError(ForbiddenOperation);
+    if (MaybeRelocatable.isFelt(other)) {
+      return new Felt((this.inner + other.inner) % Felt.PRIME);
     }
-    return new Felt((this.inner + other.inner) % Felt.PRIME);
+    throw new PrimitiveError(ForbiddenOperation);
```
zmalatrax, line 25: Throw error in the `if`condition, as in `sub()` and `div()` for consistency and readability

src/primitives/felt.ts
```diff
@@ -38,25 +38,21 @@ export class Felt {
   }
 
   mul(other: MaybeRelocatable): Felt {
-    if (!(other instanceof Felt)) {
-      throw new PrimitiveError(ForbiddenOperation);
+    if (MaybeRelocatable.isFelt(other)) {
+      return new Felt((this.inner * other.inner) % Felt.PRIME);
     }
-    return new Felt((this.inner * other.inner) % Felt.PRIME);
+    throw new PrimitiveError(ForbiddenOperation);
```
zmalatrax, line 44: Throw error in the `if`condition, as in `sub()` and `div()` for consistency and readability

src/primitives/felt.ts
tcoratger: Fixed, it was to avoid a `!` statement but sounds reasonable, just fixed.

src/primitives/felt.ts
tcoratger: Fixed, it was to avoid a `!` statement but sounds reasonable, just fixed.

src/primitives/felt.ts
```diff
@@ -19,14 +19,14 @@ export class Felt {
...
+    if (MaybeRelocatable.isFelt(other)) {
+      return new Felt((this.inner + other.inner) % Felt.PRIME);
     }
-    return new Felt((this.inner + other.inner) % Felt.PRIME);
+    throw new PrimitiveError(ForbiddenOperation);
   }
 
   sub(other: MaybeRelocatable): Felt {
-    if (!(other instanceof Felt)) {
+    if (!MaybeRelocatable.isFelt(other)) {
```
zmalatrax, line 29: The functions `isFelt()` and `isRelocatable()` are used for the class Felt in `felt.ts` but not for the class Relocatable in `relocatable.ts`, both should make use of it on methods with `MaybeRelocatable` parameters

src/primitives/relocatable.ts
```diff
@@ -8,6 +8,20 @@ import {
...
+    return maybeRelocatable instanceof Felt;
+  }
+
+  export function isRelocatable(
+    maybeRelocatable: MaybeRelocatable
+  ): maybeRelocatable is Relocatable {
+    return maybeRelocatable instanceof Relocatable;
+  }
+}
+
```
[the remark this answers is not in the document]
tcoratger: Yes I agree about creating another module, I just did it and put the two functions in it.

However, very soon we will need to implement new methods for `MaybeRelocatable` such as arithmetic operations or comparisons. So very soon function imports will become too heavy for `MaybeRelocatable` because there will be too many of them and it will be more practical to create a dedicated class in the module. But we will do that as a follow up if that suits you.

src/primitives/felt.ts
tcoratger: These assertions should all be in the same places in my opinion. Before, we had the `isFelt` inside the `Felt` class but it doesn't make much sense to put it there. It is preferable to put it in a dedicated `MaybeRelocatable` module because it is in this case of `MaybeRelocatable` that we are interested in knowing whether it is a `Felt` or a ` Relocatable`.

src/primitives/relocatable.ts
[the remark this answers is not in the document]
zmalatrax: Indeed, when the module will grow we'll refactor it
We could use wildcard import to avoid specific function imports but making a dedicated class might be a better option here

src/primitives/maybeRelocatable.ts
```diff
@@ -0,0 +1,28 @@
...
+  maybeRelocatable: MaybeRelocatable | number
+): maybeRelocatable is Felt {
+  return maybeRelocatable instanceof Felt;
+}
+
+export function isRelocatable(
+  maybeRelocatable: MaybeRelocatable
+): maybeRelocatable is Relocatable;
+export function isRelocatable(
+  maybeRelocatable: number | Relocatable
```
zmalatrax, line 22: set the union type with `number` at the end for consistency with other signatures

```suggestion
  maybeRelocatable: Relocatable | number
```

zmalatrax (approved):
lgtm

Issue #51, opened 2024-04-29
refactor: move from string types to a more typescript-semantic way


zmalatrax:
The core logic of the vm execution relies on types being union of strings that are defined in `instruction.ts`. The rationale behind it was expressiveness and readability, but it is more prone to errors (missing/replacing a character as in the test case fixed in #49) and is not 'mainstream'.

The use of enums or const object should be preferred.

PR #52, opened 2024-04-30
refactor: remove string types

zmalatrax:
Solves #51, using enums

- Rename `OpLogic` to `ResLogic`, name used in the cairo whitepaper
- Rename `RegisterFlag` to `Register` (seems ok but `RegisterFlag` might actually be more clear and should be kept)
- Cleanup of unused import/errors previously missed out
- Rename `'unconstrained'` to `ResLogic.Unused` as in the cairo whitepaper, which is easier to understand 
However, this entry might be removed in a future refactor on the operand computation (currently 'compute/deduce' vs more straightforward as in the cairo vm gs)
- Export instruction instantiation in const objects in `virtualMachine.test.ts` to reduce duplicate code and provide cleaner test cases
- Constant called from the class rather than the instance (`this.BIAS` --> `Instruction.BIAS`)
- Add JSDoc to enums with their Cairo Assembly equivalent (as in the Cairo whitepaper State transition, except for Jnz which uses a js ternary operator to show the if-else control flow)

The classes `Instruction` and `VirtualMachine` still need to be refactored but it is for another PR, as well as adding proper JSDoc to the different classes.


src/vm/instruction.ts
```diff
@@ -27,40 +27,130 @@ import {
 // └─────┴─────┴───┴───┴───┴───┴───┴───┴───┴───┴────┴────┴────┴────┴────┴────┘
 // Source: https://github.com/lambdaclass/cairo-vm_in_go/blob/main/pkg/vm/instruction.go
 
-// Dst & Op0 register flags
-export type RegisterFlag = 'ap' | 'fp';
+/** Flags to compute `dst` and `op0` from a register */
```
ClementWalter, line 30: I wouldn't put a contextual description as a comment. Here you describe the register in the context of it's usage in an instruction decoding.

Just present the register as a pure definition, and add `pc`

src/vm/instruction.ts
```diff
@@ -27,40 +27,130 @@ import {
...
+  Fp,
+}
 
-export type Op1Source = 'op0' | 'pc' | 'fp' | 'ap';
+/**
+ * Source from which op1 is computed
+ *
+ * op1 = [op1Source + op1Offset]
+ */
+export enum Op1Source {
```
ClementWalter, line 49: There is an issue regarding this #40 so maybe ok to keep this as is right not

src/vm/instruction.ts
```diff
@@ -27,40 +27,130 @@ import {
...
+  Unused,
+}
 
-export type PcUpdate =
-  | 'pc = pc'
-  | 'pc = res'
-  | 'pc = pc + res'
-  | 'res != 0 ? pc = op1 : pc += instruction_size';
+/**
+ * - Regular: pc = pc + instruction_size(Common Case)
```
ClementWalter, line 78: missing space

ClementWalter (dismissed):
I left some comments but this can be tackled later on. We need to simplify this whole instruction thing, over complicated here

src/vm/instruction.ts
zmalatrax: This enum is only used to represent the value which can be taken by `dst_reg` and `op0_reg`, which are either `ap` or `fp`, to compute the value of `dst` and `op0`. I don't think that `pc` has its place here. (See '# compute op0' and '# compute dst' in the State Transition pseudo-code of the cairo whitepaper p.33, p.34)

Knowing this context, should the register still be described with pure description (what each pointer is) or a contextual description ?

src/vm/instruction.ts
zmalatrax: Yes, I planned on tackling this issue in a next PR, while refactoring the `computeOperands()` and related methods (#54 )

src/vm/instruction.ts
ClementWalter: If the enum was to represent the value taken in the encoded instruction, then you'd have here the `0` and `1`, and not hardcoded below (which is fine for me)

```
    let dstRegister: Register;
    if (targetdstRegister == 0n) {
      dstRegister = Register.Ap;
    } else {
      dstRegister = Register.Fp;
    }
```

there is no reason for me not to have all the registers in this enum

ClementWalter (approved):

Issue #54, opened 2024-05-02
refactor: get rid of `deduce()` methods to compute operands


zmalatrax:
The current `computeOperands()` method is based on the production vm in Rust:
- Compute the operand address
- Read the value at the computed address
- If the read memory cell is empty, deduce the operand value from the other operands with `deduce()`

Burying this logic in multiple methods with their own switch-case flow is confusing. It can, and should be done in a simpler way, straightforwardly.

PR #57, opened 2024-05-14
refactor: decode instruction & compute operands

zmalatrax:
Solves #54, #40

- Rework of the `Register` type and removal of `Op1Source`: `op0`, `dst` & `op1` should be handled the same way as registers. (op1Source might be renamed top op1Register for clarity) 
- Instruction decoding has been cleaned up to compute the different flags and enforce the use of switch cases for each flag.
- Move away from the `computeX()`/`deduceX()` methods from the Python andRust VMs. Use a `switch` on two variables instead (opcode and resLogic, as Cairo instructions are like `opcode(dst, res)` with `res = res_logic(op0, op1)`), suggested by @ClementWalter


src/vm/virtualMachine.ts
```diff
@@ -150,224 +108,192 @@ export class VirtualMachine {
...
+    let op0: SegmentValue | undefined;
+    let op1: SegmentValue | undefined;
 
-      if (deducedOp1 !== undefined) {
-        this.memory.assertEq(op1Addr, deducedOp1);
-      }
+    // Compute the destination address based
+    // on the dstRegister and the offset
+    // Example: const dstAddr = this.fp.add(5) == fp + 5;
+    switch (dstRegister) {
```
ClementWalter, line 134:
```
dstAddr = this.registers[dstRegister].add(dstOffset)
// etc.
```

and throw errors while decoding the instruction

src/vm/virtualMachine.ts
```diff
@@ -150,224 +108,192 @@ export class VirtualMachine {
...
+        break;
+      default:
+        throw new InvalidOp0Register();
+    }
+    op0 = this.memory.get(op0Addr);
 
-      res = computedRes;
+    // Compute the second operand address based
+    // on the op1Source and the offset
+    switch (op1Source) {
```
ClementWalter, line 162: op1Register

src/vm/virtualMachine.ts
zmalatrax: To do so, we need to define a mapping from the Register enum values to the corresponding attribute. 
As it is only used in the `computeOperands()` method (renamed `computeAuxValues()` in more recent commits), defining it as a variable makes more sense than defining an attribute imo

```typescript
computeAuxValues(instruction: Instruction) {
  // [...]

  const registers = {
    [Register.Pc]: this.pc,
    [Register.Ap]: this.ap,
    [Register.Fp]: this.fp,
  };

  const dstAddr = registers[dstRegister].add(dstOffset);
  
// [...]
}
```

The errors thrown in the default switch cases were actually unnecessary as already handled when decoding instructions

ClementWalter: what about removing the `this.pc`, `this.ap` and `this.pc` and only have `this.registers`?

src/vm/virtualMachine.ts
ClementWalter, line 1:
```
  step(): void {
    const maybeEncodedInstruction = this.memory.get(this.pc);
    if (maybeEncodedInstruction === undefined) {
      throw new EndOfInstructionsError();
    }
    if (!isFelt(maybeEncodedInstruction)) {
      throw new InstructionError();
    }
```

1. I don't understand why we raise "End of" while it's more like "Undefined instruction"
2. why this check about instruction being a felt?

src/vm/virtualMachine.ts
```diff
@@ -125,381 +57,292 @@ export class VirtualMachine {
 
     const encodedInstruction = maybeEncodedInstruction.toBigInt();
 
-    // decode and run instruction
     const instruction = Instruction.decodeInstruction(encodedInstruction);
 
-    return this.runInstruction(instruction);
+    this.runInstruction(instruction);
```
ClementWalter, line 62: so `step` is basically `decode + run` and ` decode` belongs to `Instruction`. I'd then inline the `run` here

src/vm/virtualMachine.ts
```diff
@@ -125,381 +57,292 @@ export class VirtualMachine {
...
+   *
+   * "They can be computed from the memory values,
+   * the three offsets, and the instruction flags."
+   */
+  computeStepValues(instruction: Instruction): {
+    op0: SegmentValue | undefined;
+    op1: SegmentValue | undefined;
+    res: SegmentValue | undefined;
+    dst: SegmentValue | undefined;
+  } {
```
ClementWalter, line 95: why not creating an explicit type for these `StepValues`

src/vm/virtualMachine.ts
```diff
@@ -125,381 +57,292 @@ export class VirtualMachine {
...
-      default:
-        return undefined;
-    }
+    op0: SegmentValue | undefined,
+    op1: SegmentValue | undefined
+  ): { op0: SegmentValue; op1: SegmentValue } {
+    const nextPc = this.pc.add(instruction.size());
+    if (op0 === undefined) {
+      op0 = nextPc;
+    } else if (!op0.eq(nextPc)) throw new InvalidOp0();
```
ClementWalter, line 238: this could be made more explicit that the validation is about 

1. op0 === nextPc
2. op1 !== undefined

so I'd move away the op0 resolution

```
op0 = op0 === undefined ? nextPc : op0;
```

possibly compute nextPc, nextAp and nextFp before so that it's done and here
and write only the `if then throw` that could possibly be inlined as well

ClementWalter (changes requested):

ClementWalter (approved):

src/vm/virtualMachine.ts
```diff
@@ -125,381 +56,292 @@ export class VirtualMachine {
...
+        dst = this.fp;
+        break;
 
-      if (deducedDst !== undefined) {
-        this.memory.assertEq(dstAddr, deducedDst);
-      }
-      dst = deducedDst;
-    }
+      case Opcode.AssertEq | ResLogic.Op1:
+        if (op1 === undefined) op1 = dst;
```
ClementWalter, line 148: maybe `op1 = op1 ?? dst`

src/vm/virtualMachine.ts
```diff
@@ -125,381 +56,292 @@ export class VirtualMachine {
...
+        res = op0.mul(op1);
+        dst = this.fp;
+        break;
 
-    // If dst is undefined, then we can deduce it from the instruction and res
-    if (dst === undefined) {
-      const deducedDst = this.deduceDst(instruction, res);
+      case Opcode.Call | ResLogic.Unused:
+        ({ op0, op1 } = this.validateOperandsCallOpcode(instruction, op0, op1));
+        res = undefined;
```
ClementWalter, line 143: `res` is already `undefined` no?

src/vm/virtualMachine.ts
```diff
@@ -125,381 +56,292 @@ export class VirtualMachine {
...
+        if (op0 === undefined) {
+          if (dst !== undefined && op1 !== undefined) {
+            op0 = dst.sub(op1);
+          } else throw new InvalidOp0();
+        }
+        if (op1 === undefined) {
+          if (dst !== undefined && op0 !== undefined) {
+            op1 = dst.sub(op0);
+          } else throw new InvalidOp1();
+        }
```
ClementWalter, line 163:
```
if (op0 === undefined) {
    if (dst === undefined ||op1 == undefined) {
        throw new InvalidOp0();
    }
    op0 = dst.sub(op1);
}

if (op1 === undefined) {
    if (dst === undefined || op0 === undefined) {
        throw new InvalidOp1();
    }
    op1 = dst.sub(op0);
}
```

etc.

src/vm/virtualMachine.ts
zmalatrax: `step` will later execute hints before handling the instruction, we should keep it as is

src/vm/virtualMachine.ts
zmalatrax: Indeed, same goes for the other case with ResLogic.Unused (in the default case). I wanted to be explicit but as the switch is directly after the declaration of res (to undefined), it is useless --> to be removed

src/vm/virtualMachine.ts
```diff
@@ -125,381 +56,297 @@ export class VirtualMachine {
...
+        break;
     }
-  }
 
-  // Update the registers based on the instruction.
-  updateRegisters(instruction: Instruction, operands: Operands): void {
-    this.updatePc(instruction, operands);
+    if (op0 !== undefined) this.memory.assertEq(op0Addr, op0);
+    if (op1 !== undefined) this.memory.assertEq(op1Addr, op1);
+    if (dst !== undefined) this.memory.assertEq(dstAddr, dst);
```
ClementWalter, line 237: how could they be `undefined`?

ClementWalter (approved):

Issue #43, opened 2023-12-16
feat: implement relocate 


Eikix:
based on https://github.com/lambdaclass/cairo-vm_in_go?tab=readme-ov-file#relocate

implement relocate

henryjindal112:
@Eikix, Please assign this to me

Issue #32, opened 2023-11-16
feat: generate the trace of execution 


greged93:
Generate the trace of a cairo execution. Include in the CI a way to run and generate the trace with another battle tested VM (such as the python or rust cairo VM) and compare both generated traces.

zmalatrax:
The `trace` has been implemented, still needs to handle diff testing with another VM and integrate it in the CI.

Re-scoped in #59

PR #58, opened 2024-05-17
feat: implement trace and relocate

zmalatrax:
Solves #43 and partially #32.

- Relocate memory. The relocated memory starts at address 1 (and not 0), a Map is then used for readability
- Trace. The trace is simply a table appending the value of each registers at every step.
The type `TraceEntry` is an object containing the three registers, similar to what would be the type `Registers` if we had an attribute `registers` in the VM (potential refactoring of the attributes `pc`, `ap` and `fp`)
- Relocate Trace (not implemented yet)
- Encode Memory (resp. Trace) and write it to a file (not implemented yet)

Having a CI job that generates the execution traces (and memory) of the Cairo VM in TS + another one (or multiple ones?) such as rust or python (prod-ready) and compare them would be great. But it still needs to be thought upon and it isn't not a priority: a separate PR should be done later.

### DIFF
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -59,11 +59,16 @@ export class Memory {
 
   toString(): string {
     const buffer = ['\nMEMORY', 'Address  ->  Value', '-----------------'];
-    for (const [index, segment] of this.values.entries()) {
-      for (const [offset, value] of segment.entries()) {
-        buffer.push(`${index}:${offset} -> ${value.toString()}`);
-      }
-    }
+    buffer.push(
+      this.values
+        .map((segment, index) =>
+          segment.map(
+            (value, offset) => `${index}:${offset} -> ${value.toString()}`
+          )
+        )
+        .flat()
+        .join('\n')
+    );
     return buffer.join('\n');
   }
 }
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -57,14 +57,17 @@ export class Memory {
     return this.values[segment]?.length ?? 0;
   }
 
-  printMemory() {
-    console.log('MEMORY');
-    console.log('Address  |  Value');
-    console.log('-----------------');
+  toString(): string {
+    const memoryToPrint = [
+      '\nMEMORY',
+      'Address  ->  Value',
+      '-----------------',
+    ];
     for (const [index, segment] of Object.entries(this.values)) {
       for (const [offset, value] of segment.entries()) {
-        console.log(`${index}:${offset} -> ${value.toString()}`);
+        memoryToPrint.push(`${index}:${offset} -> ${value.toString()}`);
       }
     }
+    return memoryToPrint.join('\n');
   }
 }
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -43,11 +43,8 @@ export class Memory {
       throw new SegmentOutOfBounds(segment, segmentNumber);
     }
 
-    const currentValue: SegmentValue | undefined = this.values[segment][offset];
-
-    if (currentValue === undefined) {
-      this.values[segment][offset] = value;
-    } else if (currentValue !== value) {
+    this.values[segment][offset] = this.values[segment][offset] ?? value;
+    if (this.values[segment][offset] !== value) {
       throw new InconsistentMemory(
         address,
         this.values[segment][offset],
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -58,16 +58,12 @@ export class Memory {
   }
 
   toString(): string {
-    const memoryToPrint = [
-      '\nMEMORY',
-      'Address  ->  Value',
-      '-----------------',
-    ];
+    const buffer = ['\nMEMORY', 'Address  ->  Value', '-----------------'];
     for (const [index, segment] of this.values.entries()) {
       for (const [offset, value] of segment.entries()) {
-        memoryToPrint.push(`${index}:${offset} -> ${value.toString()}`);
+        buffer.push(`${index}:${offset} -> ${value.toString()}`);
       }
     }
-    return memoryToPrint.join('\n');
+    return buffer.join('\n');
   }
 }
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -59,4 +59,15 @@ export class Memory {
   getSegmentSize(segment: number): number {
     return this.values[segment]?.length ?? 0;
   }
+
+  printMemory() {
+    console.log('MEMORY');
+    console.log('Address  |  Value');
+    console.log('-----------------');
+    for (const [index, segment] of Object.entries(this.values)) {
+      for (const [offset, value] of segment.entries()) {
+        console.log(`${index}:${offset} -> ${value.toString()}`);
+      }
+    }
+  }
 }
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -63,7 +63,7 @@ export class Memory {
       'Address  ->  Value',
       '-----------------',
     ];
-    for (const [index, segment] of Object.entries(this.values)) {
+    for (const [index, segment] of this.values.entries()) {
       for (const [offset, value] of segment.entries()) {
         memoryToPrint.push(`${index}:${offset} -> ${value.toString()}`);
       }
--- a/src/memory/memory.ts
+++ b/src/memory/memory.ts
@@ -58,17 +58,17 @@ export class Memory {
   }
 
   toString(): string {
-    const buffer = ['\nMEMORY', 'Address  ->  Value', '-----------------'];
-    buffer.push(
-      this.values
-        .map((segment, index) =>
-          segment.map(
-            (value, offset) => `${index}:${offset} -> ${value.toString()}`
-          )
+    return [
+      '\nMEMORY',
+      'Address  ->  Value',
+      '-----------------',
+      ...this.values.map((segment, index) =>
+        segment.map(
+          (value, offset) => `${index}:${offset} -> ${value.toString()}`
         )
-        .flat()
-        .join('\n')
-    );
-    return buffer.join('\n');
+      ),
+    ]
+      .flat()
+      .join('\n');
   }
 }
--- a/src/runners/cairoRunner.test.ts
+++ b/src/runners/cairoRunner.test.ts
@@ -30,8 +30,7 @@ describe('cairoRunner', () => {
   describe('runUntilPc', () => {
     test('should return the value of the 10th fibonacci number', () => {
       const runner = new CairoRunner(PROGRAM);
-      const finalPc = new Relocatable(0, 12);
-      runner.runUntilPc(finalPc, true, true);
+      runner.runUntilPc(runner.finalPc, true, true);
       const executionSize = runner.vm.memory.getSegmentSize(1);
       const executionEnd = runner.executionBase.add(executionSize);
 
@@ -40,8 +39,8 @@ describe('cairoRunner', () => {
 
     test('should export encoded trace and memory', () => {
       const runner = new CairoRunner(PROGRAM);
-      const finalPc = new Relocatable(0, 12);
-      runner.runUntilPc(finalPc, true, true);
+      // const finalPc = new Relocatable(0, 12);
+      runner.runUntilPc(runner.finalPc, true, true);
       const executionSize = runner.vm.memory.getSegmentSize(1);
       const executionEnd = runner.executionBase.add(executionSize);
       expect(runner.vm.memory.get(executionEnd.sub(1))).toEqual(new Felt(144n));
--- a/src/runners/cairoRunner.test.ts
+++ b/src/runners/cairoRunner.test.ts
@@ -34,7 +34,7 @@ describe('cairoRunner', () => {
   describe('runUntilPc', () => {
     test('should return the value of the 10th fibonacci number', () => {
       const runner = new CairoRunner(PROGRAM);
-      runner.runUntilPc(runner.finalPc, false, true, true);
+      runner.runUntilPc(runner.finalPc, true, 0);
       const executionSize = runner.vm.memory.getSegmentSize(1);
       const executionEnd = runner.executionBase.add(executionSize);
 
@@ -50,7 +50,7 @@ describe('cairoRunner', () => {
     */
     test('should export encoded trace', () => {
       const runner = new CairoRunner(PROGRAM);
-      runner.runUntilPc(runner.finalPc, true);
+      runner.runUntilPc(runner.finalPc, true, 1);
       const trace_filename = 'fibonacci_trace_ts.bin';
       const trace_path = path.join(tmpDir, trace_filename);
       runner.exportTrace(trace_path);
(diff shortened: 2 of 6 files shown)


src/memory/memory.ts
```diff
@@ -59,4 +56,15 @@ export class Memory {
   getSegmentSize(segment: number): number {
     return this.values[segment]?.length ?? 0;
   }
+
+  printMemory() {
```
ClementWalter, line 60: isn't it more idiomatic to call this `toString()`, return a `string` and eventually use it like

```
console.log(memory.toString())
```

?

zmalatrax: Indeed, much better this way
Made something similar for printing `relocatedMemory`

src/vm/virtualMachine.ts
```diff
@@ -351,4 +361,42 @@ export class VirtualMachine {
...
+
+    for (const segment of this.memory.values) {
+      memorySize += segment.length;
+      relocationTable.push(memorySize);
+    }
+
+    for (const [index, segment] of Object.entries(this.memory.values)) {
+      for (const [offset, value] of segment.entries()) {
+        let relocatedValue: Felt;
+        const relocatedAddr = relocationTable[Number(index)] + Number(offset);
```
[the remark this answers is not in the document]
zmalatrax: They are

Dunno why but I've used `Object.entries(this.memory.values)` (which returns a string for the key) instead of the array method `this.memory.values.entries()` to get `index` and `segment`...

src/memory/memory.ts
```diff
@@ -59,4 +56,18 @@ export class Memory {
   getSegmentSize(segment: number): number {
     return this.values[segment]?.length ?? 0;
   }
+
+  toString(): string {
+    const memoryToPrint = [
```
ClementWalter, line 61: I'd rather call this `buffer` or `output`

src/memory/memory.ts
```diff
@@ -59,4 +56,18 @@ export class Memory {
...
+      '\nMEMORY',
+      'Address  ->  Value',
+      '-----------------',
+    ];
+    for (const [index, segment] of this.values.entries()) {
+      for (const [offset, value] of segment.entries()) {
+        memoryToPrint.push(`${index}:${offset} -> ${value.toString()}`);
+      }
+    }
+    return memoryToPrint.join('\n');
```
ClementWalter, line 71: I may be biased but in the case of a "pure" interation, I think that map is 100x better than for loop to explicitly show that you don't mix values from one interation to another, but just apply the same thing to each item

```
this.values
    .map((segment, index) =>
    segment.map(
        (value, offset) => `${index}:${offset} -> ${value.toString()}`
    )
    )
    .flat()
    .join('\n');
```

src/runners/cairoRunner.test.ts
```diff
@@ -30,11 +31,26 @@ describe('cairoRunner', () => {
...
+      runner.runUntilPc(finalPc, false, true, true);
       const executionSize = runner.vm.memory.getSegmentSize(1);
       const executionEnd = runner.executionBase.add(executionSize);
 
       expect(runner.vm.memory.get(executionEnd.sub(1))).toEqual(new Felt(144n));
     });
+
+    test('should export encoded trace and memory', () => {
+      const runner = new CairoRunner(PROGRAM);
+      const finalPc = new Relocatable(0, 12);
```
ClementWalter, line 43: I see that you copy/pasted this from the test above, but you really don't want to do this because it's program dependant. What if your test program gets longer and 12 is not the first empty slot anymore?

In the python vm, it's just to random relocatable, got by creating new segments

src/memory/memory.ts
zmalatrax: Indeed, much cleaner

src/memory/memory.ts
```diff
@@ -59,4 +56,19 @@ export class Memory {
...
+    buffer.push(
+      this.values
+        .map((segment, index) =>
+          segment.map(
+            (value, offset) => `${index}:${offset} -> ${value.toString()}`
+          )
+        )
+        .flat()
+        .join('\n')
+    );
```
ClementWalter, line 71: what about 

```
return [
    'MEMORY',
    'Address  ->  Value', '-----------------',
    ...this.values
        .map((segment, index) =>
          segment.map(
            (value, offset) => `${index}:${offset} -> ${value.toString()}`
          )
        )
].flat().join('\n')
```

src/runners/cairoRunner.test.ts
zmalatrax: Yep, made a dummy copy/paste to quickly check that the encoding was working
Declaring a new variable `finalPc` instead of using the one initialized in the runner, `runner.finalPc` doesn't make sense anyway

Some more refactoring might be done in a later PR

src/runners/cairoRunner.test.ts
```diff
@@ -29,12 +30,26 @@ describe('cairoRunner', () => {
...
+      runner.runUntilPc(runner.finalPc, true, true);
       const executionSize = runner.vm.memory.getSegmentSize(1);
       const executionEnd = runner.executionBase.add(executionSize);
 
       expect(runner.vm.memory.get(executionEnd.sub(1))).toEqual(new Felt(144n));
     });
+
+    test('should export encoded trace and memory', () => {
+      const runner = new CairoRunner(PROGRAM);
+      // const finalPc = new Relocatable(0, 12);
```
ClementWalter, line 42: remove commented line

src/runners/cairoRunner.test.ts
```diff
@@ -29,12 +30,26 @@ describe('cairoRunner', () => {
...
+      runner.runUntilPc(runner.finalPc, true, true);
+      const executionSize = runner.vm.memory.getSegmentSize(1);
+      const executionEnd = runner.executionBase.add(executionSize);
+      expect(runner.vm.memory.get(executionEnd.sub(1))).toEqual(new Felt(144n));
```
ClementWalter, line 46: don't assert in this **unit** test the predicates of other tests

src/runners/cairoRunner.test.ts
```diff
@@ -29,12 +30,26 @@ describe('cairoRunner', () => {
...
+
+      const dir = './cairo_programs/cairo_0';
```
ClementWalter, line 48: use tmp dir

src/vm/virtualMachine.ts
```diff
@@ -351,4 +369,58 @@ export class VirtualMachine {
...
+
+  /** Relocate memory and trace */
+  relocate() {
+    let relocationTable = [1];
+    let memorySize = 1;
+
+    for (const segment of this.memory.values) {
+      memorySize += segment.length;
+      relocationTable.push(memorySize);
+    }
```
ClementWalter, line 381: so your `relocatationTable` is the start index of the segment in the final flatten memory

hence you want to do instead of cumsum of the segment lengths

```
this.memory.values
    .map((segment) => segment.length)
    .map((sum => value => sum += value)(0))
```

src/vm/virtualMachine.ts
```diff
@@ -351,4 +369,58 @@ export class VirtualMachine {
...
+        } else if (isRelocatable(value)) {
+          relocatedValue = new Felt(
+            BigInt(relocationTable[value.segment] + value.offset)
+          );
+        } else {
+          throw new Error();
+        }
+
+        this.relocatedMemory.set(relocatedAddr, relocatedValue);
+      }
```
ClementWalter, line 398: then this is easier formulated as a simple `.flat` with a `.map` over each value to convert `Relocatable`

```
this.memory.values.flat().map((value) => isFelt(value) ? value : new Felt(
            BigInt(relocationTable[value.segment] + value.offset)
          )
```

src/vm/virtualMachine.ts
```diff
@@ -351,4 +369,58 @@ export class VirtualMachine {
...
+        let relocatedValue: Felt;
+        const relocatedAddr = relocationTable[index] + offset;
+        if (isFelt(value)) {
+          relocatedValue = value;
+        } else if (isRelocatable(value)) {
+          relocatedValue = new Felt(
+            BigInt(relocationTable[value.segment] + value.offset)
+          );
+        } else {
+          throw new Error();
```
ClementWalter, line 394: memory holes don't generate error afaik

src/vm/virtualMachine.ts
```diff
@@ -351,4 +369,58 @@ export class VirtualMachine {
...
+        }
+
+        this.relocatedMemory.set(relocatedAddr, relocatedValue);
```
ClementWalter, line 397: do we really need an stateful attribute for this?

src/vm/virtualMachine.ts
```diff
@@ -351,4 +369,58 @@ export class VirtualMachine {
...
+      }
+    }
+    for (const entry of this.trace) {
```
ClementWalter, line 400:
```
this.trace.map(({pc, ap, fp}) => ...
```

not sure where and how this is used but could be as well

```
this.trace.map(Object.values).flat().map((register) => new Felt(
          BigInt(relocationTable[entry.pc.segment] + entry.pc.offset)
        )
```

src/vm/virtualMachine.ts
```diff
@@ -351,4 +369,58 @@ export class VirtualMachine {
...
+    const buffer = [
+      '\nRELOCATED MEMORY',
+      'Address  ->  Value',
+      '-----------------',
+    ];
+    for (const [index, value] of this.relocatedMemory.entries()) {
+      buffer.push(`${index} -> ${value.toString()}`);
+    }
+    return buffer.join('\n');
+  }
```
ClementWalter, line 425: refacto with `.map`

src/runners/cairoRunner.ts
```diff
@@ -47,17 +49,53 @@ export class CairoRunner {
...
+
+    this.vm.relocatedMemory.forEach((value, address) => {
+      const byteOffset = (address - 1) * 5 * 8;
+      view.setBigUint64(byteOffset, BigInt(address), true);
+      view.setBigUint64(byteOffset + 8, value.toBigInt(), true);
+    });
+
+    fs.writeFile(filename, buffer, { flag: 'w+' }, (err) => {
+      if (err) throw err;
+    });
```
ClementWalter, line 99: refacto with `.map`

src/runners/cairoRunner.ts
```diff
@@ -47,17 +49,53 @@ export class CairoRunner {
...
+  }
+
+  /** Export the relocated memory little-endian encoded to a file */
+  exportMemory(filename: string = 'encoded_memory') {
+    const buffer = new ArrayBuffer(this.vm.relocatedMemory.size * 5 * 8);
+    const view = new DataView(buffer);
+
+    this.vm.relocatedMemory.forEach((value, address) => {
+      const byteOffset = (address - 1) * 5 * 8;
```
ClementWalter, line 92: what is this `*5` ?

src/vm/virtualMachine.ts
zmalatrax: Yes, so `relocationTable` must starts with the offset of the segment 0, which is 1 (the relocated memory is 1-based...)

Mapping each segment to its length and directly do the cumulative sum will not include the offset of segment 0 and the correct offset for a 1-based memory

To use `.map` and have the correct relocation table we can do this:
```typescript
const relocationTable = this.memory.values
  .map((segment) => segment.length)
  .map(
    ((sum) => (value, index) =>
      index ? (sum += value) : [1, (sum += value)])(1)
  )
  .flat();
```

Or this, but I prefer the first version:

```typescript
const relocationTable = [
  1,
  ...this.memory.values
    .map((segment) => segment.length)
    .map(
      (
        (sum) => (value) =>
          (sum += value)
      )(1)
    ),
];
```

zmalatrax: Actually the offset of segment 0 was included but miscalculated (adding its length to its final base address when it shouldn't)

A working solution would be:
```typescript
const relocationTable = this.memory.values
  .map((segment) => segment.length)
  .map(
    (
      (sum) => (value) =>
        (sum += value) - value
    )(1)
  );
```

src/runners/cairoRunner.ts
zmalatrax: The address is written on 1 byte while the value (being a Felt) is written on 4 bytes, thus the `* 5`

src/runners/cairoRunner.test.ts
```diff
@@ -29,12 +30,24 @@ describe('cairoRunner', () => {
...
       const executionSize = runner.vm.memory.getSegmentSize(1);
       const executionEnd = runner.executionBase.add(executionSize);
 
       expect(runner.vm.memory.get(executionEnd.sub(1))).toEqual(new Felt(144n));
     });
+
+    test('should export encoded trace and memory', () => {
+      const runner = new CairoRunner(PROGRAM);
+      runner.runUntilPc(runner.finalPc, true, true);
+      const dir = 'tmp';
```
ClementWalter, line 43: use a true tmp dir

src/runners/cairoRunner.test.ts
```diff
@@ -29,12 +34,20 @@ describe('cairoRunner', () => {
...
     });
+
+    test('should export encoded trace and memory', () => {
+      const runner = new CairoRunner(PROGRAM);
+      runner.runUntilPc(runner.finalPc, true, true);
+      const trace_filename = 'fibonacci_trace_ts.bin';
+      const memory_filename = 'fibonacci_memory_ts.bin';
+      runner.exportTrace(path.join(tmpDir, trace_filename));
+      runner.exportMemory(path.join(tmpDir, memory_filename));
+    });
```
ClementWalter, line 51: so this is 2 tests in one 1 test, plz split

+, you should at least assert that the file exists, and better also test the content of the file. We could run this test with a simpler cairo program ?

src/vm/virtualMachine.ts
```diff
@@ -351,4 +374,44 @@ export class VirtualMachine {
...
+  /** Relocate memory and trace */
+  relocate() {
+    const relocationTable = this.memory.values
+      .map((segment) => segment.length)
+      .map(
+        (
+          (sum) => (value) =>
+            (sum += value) - value
+        )(1)
+      );
```
ClementWalter, line 387: if you want to keep this weird thing, plz add a note and a todo

src/vm/virtualMachine.ts
```diff
@@ -351,4 +374,44 @@ export class VirtualMachine {
...
+    this.relocatedMemory = this.memory.values.flatMap((segment, index) =>
+      segment.map((value, offset) => ({
+        address: relocationTable[index] + offset,
+        value: isFelt(value)
+          ? value
+          : new Felt(BigInt(relocationTable[value.segment] + value.offset)),
+      }))
+    );
+
+    this.relocatedTrace = this.trace.map(({ pc, ap, fp }) => ({
```
ClementWalter, line 398:
```
this.trace.map(Object.values).map((register) => new Felt(BigInt(relocatationTable[register.segment] + register.offset))
```

src/vm/virtualMachine.ts
```diff
@@ -23,16 +23,39 @@ import {
...
+  trace: TraceEntry[];
+  relocatedTrace: RelocatedTraceEntry[];
+  relocatedMemory: RelocatedMemory[];
 
   constructor() {
     this.currentStep = 0n;
     this.memory = new Memory();
+    this.relocatedMemory = [];
+    this.trace = [];
+    this.relocatedTrace = [];
```
ClementWalter, line 58: image
plz decide an order between memory/trace/relocatedM/relocatedTrace and memory/relocatedM/trace/relocatedT and keep

ClementWalter (approved):

PR #61, opened 2024-05-24
refactor: Cairo Runner

zmalatrax:
Refactor the Cairo Runner to simplify it:

- Remove unnecessary attributes
- Cleanup constructor
- Group configuration options into a single object, might be expanded with features to be implemented

The method `runUntilPc` is renamed to `run` because the value of `pc` to run up to is computed in the constructor (`finalPc` attribute). Having the relocation of memory and trace is  much cleaner

- Refactor memory constructor to initialize memory with no segments
- Add test for printing methods of memory & relocated memory

### DIFF
--- a/src/memory/memory.test.ts
+++ b/src/memory/memory.test.ts
@@ -1,58 +1,66 @@
 import { test, expect, describe } from 'bun:test';
 
 import { SegmentOutOfBounds, InconsistentMemory } from 'errors/memory';
-
 import { Felt } from 'primitives/felt';
 import { Relocatable } from 'primitives/relocatable';
 import { Memory } from './memory';
 
+const VALUES = [
+  new Relocatable(0, 0),
+  new Relocatable(0, 1),
+  new Felt(1n),
+  new Felt(2n),
+  new Relocatable(1, 1),
+];
+
 describe('Memory', () => {
+  describe('constructor', () => {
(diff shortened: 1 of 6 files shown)


src/vm/virtualMachine.test.ts
```diff
@@ -502,4 +502,39 @@ describe('VirtualMachine', () => {
...
+        'Address  ->  Value',
+        '-----------------',
+        `0 -> 1`,
+        `2 -> 2`,
+        `3 -> 2`,
+      ].join('\n');
+
+      const logSpy = spyOn(vm, 'relocatedMemoryToString');
+
+      console.log(vm.relocatedMemoryToString());
```
ClementWalter, line 535: remove

src/memory/memory.test.ts
```diff
@@ -95,4 +103,35 @@ describe('Memory', () => {
...
+        'Address  ->  Value',
+        '-----------------',
+        '0:0 -> 1',
+        '0:2 -> 2',
+        '1:0 -> 0:1',
+      ].join('\n');
+
+      const logSpy = spyOn(memory, 'toString');
+
+      console.log(memory.toString());
```
ClementWalter, line 132: remove

ClementWalter (approved):

PR #62, opened 2024-05-27
feat: builtin runner

zmalatrax:
- Add builtin runner
- Add Bitwise builtin
- Refactor memory, introducing Segment class to handle builtins

### DIFF
--- a/cairo_programs/cairo_0/bitwise_builtin_test.cairo
+++ b/cairo_programs/cairo_0/bitwise_builtin_test.cairo
@@ -1,0 +1,20 @@
+// Taken from lambda-class/cairo-vm repo
+
+%builtins bitwise
+from starkware.cairo.common.bitwise import bitwise_and, bitwise_xor, bitwise_or, bitwise_operations
+from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
+
+func main{bitwise_ptr: BitwiseBuiltin*}() {
+    let (and_a) = bitwise_and(12, 10);  // Binary (1100, 1010).
+    assert and_a = 8;  // Binary 1000.
+    let (xor_a) = bitwise_xor(12, 10);
+    assert xor_a = 6;
+    let (or_a) = bitwise_or(12, 10);
+    assert or_a = 14;
+
+    let (and_b, xor_b, or_b) = bitwise_operations(9, 3);
+    assert and_b = 1;
+    assert xor_b = 10;
+    assert or_b = 11;
+    return ();
+}
--- a/src/builtins/builtin.ts
+++ b/src/builtins/builtin.ts
@@ -16,8 +16,11 @@ export type BuiltinHandler = ProxyHandler<Array<SegmentValue>>;
  * - Segment Arena: Unknown usage, must investigate
  * - Output: Output builtin
  */
-export const BUILTIN_HANDLER: {
+const BUILTIN_HANDLER: {
   [key: string]: BuiltinHandler;
 } = {
   bitwise: bitwiseHandler,
 };
+
+/** Getter of the object `BUILTIN_HANDLER` */
+export const getBuiltin = (name: string) => BUILTIN_HANDLER[name];
--- a/src/builtins/builtin.ts
+++ b/src/builtins/builtin.ts
@@ -1,24 +1,11 @@
-import { Segment } from 'memory/memory';
-import { Relocatable } from 'primitives/relocatable';
 import { SegmentValue } from 'primitives/segmentValue';
+import { bitwiseHandler } from './bitwise';
 
-/** Abstract class representing builtin runners */
-export abstract class BuiltinRunnerBase {
-  constructor() {}
-  /** Verify that a property of the segment (if any) is respected
-   * when asserting constraints (`assertEq`) */
-  validate(_segment: Segment): void {}
-  /** Infer value from the previous cells of the segment when reading from memory (`get`)*/
-  infer(_segment: Segment, _address: Relocatable): SegmentValue | undefined {
-    return undefined;
-  }
-  /** @return builtin name */
-  abstract toString(): string;
-}
+/** Proxy handler to abstract validation & deduction rules off the VM */
+export type BuiltinHandler = ProxyHandler<Array<SegmentValue>>;
 
 /**
- * Union of the different builtin runners:
- * - NoBuiltin: memory segments that are not builtins
+ * Object to map builtin names to their ProxyHandler:
  * - Bitwise: Builtin for bitwise operations
  * - RangeCheck: Builtin for inequality operations
  * - ECDSA: Builtin for Elliptic Curve Digital Signature Algorithm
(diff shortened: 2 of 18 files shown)


src/vm/virtualMachine.ts
ClementWalter, line 1: looks like you don't have trunk running in VS Code
image

src/runners/cairoRunner.ts
```diff
@@ -30,14 +31,23 @@ export class CairoRunner {
...
+        case 'bitwise':
+          return this.vm.memory.addSegment(new Bitwise());
+        default:
+          throw new Error();
+      }
+    });
     const returnFp = this.vm.memory.addSegment();
     this.finalPc = this.vm.memory.addSegment();
-    const stack = [returnFp, this.finalPc];
+    const stack = [...builtin_stack, returnFp, this.finalPc];
```
ClementWalter, line 50: builtins have an initial_stack method, they may not all just add a pointer in the stack (see python VM)

src/runners/cairoRunner.ts
```diff
@@ -30,14 +31,23 @@ export class CairoRunner {
...
 
     this.vm = new VirtualMachine();
     this.programBase = this.vm.memory.addSegment();
     this.executionBase = this.vm.memory.addSegment();
 
+    const builtin_stack = builtins.map((builtin) => {
+      switch (builtin) {
+        case 'bitwise':
+          return this.vm.memory.addSegment(new Bitwise());
+        default:
```
ClementWalter, line 44: I'd rather export an object/namespace/class from builtins so that here we can do only `program.builtins.map((b) => new Builtin(b)).map((b) => b.initial_stack()).flat()`

ClementWalter: especially here at the cairoRunner level, I don't really want to have this detailed list of what builtin name corresponds to what builtin class, but I'd rather export an object so that the module builtin is self-sufficient

src/memory/memory.ts
```diff
@@ -2,29 +2,54 @@ import { InconsistentMemory, SegmentOutOfBounds } from 'errors/memory';
 
 import { Relocatable } from 'primitives/relocatable';
 import { SegmentValue } from 'primitives/segmentValue';
+import { BuiltinRunner, NoBuiltin } from 'builtins/builtin';
+
+export class Segment {
```
ClementWalter, line 7: why not extending `Array<SegmentValue>` ? see also https://stackoverflow.com/questions/2449182/getter-setter-on-javascript-array to update the getter for given builtins

src/errors/builtins.ts
ClementWalter, line 1: see builtins comments, I think that builtins are just `Segments`

src/builtins/builtin.ts
```diff
@@ -0,0 +1,48 @@
+import { Segment } from 'memory/memory';
+import { Relocatable } from 'primitives/relocatable';
+import { SegmentValue } from 'primitives/segmentValue';
+
+/** Abstract class representing builtin runners */
+export abstract class BuiltinRunnerBase {
```
ClementWalter, line 6: I don't really understand why we need a dedicated builtin runner. It seems to me that we can just add rules to segments getter, default segments being non-builtins segments just return the value of the underlying array, and builtins segments apply the rule their. Consequently, from the VM run point of view, builtins are transparent

src/memory/memory.ts
```diff
@@ -2,29 +2,54 @@ import { InconsistentMemory, SegmentOutOfBounds } from 'errors/memory';
...
+  builtin: BuiltinRunner;
+  values: Array<SegmentValue>;
+
+  constructor(builtin: BuiltinRunner = new NoBuiltin()) {
+    this.builtin = builtin;
+    this.values = [];
+  }
+
+  get(address: Relocatable): SegmentValue | undefined {
```
ClementWalter, line 16: `get` should only have an offset and not a `Relocatable` (what would this mean if the segment is not the same?)

src/memory/memory.ts
zmalatrax: As discussed in a call, segments are refactored in [Proxy ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) of `Array<SegmentValue>`, builtins having a dedicated handler modifying the `[[Set]]` and/or `[[Get]]` methods to enforce their properties

src/errors/builtins.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/** Value cannot be infered from undefined cell value */
+export class UndefinedValue extends BuiltinError {
+  public readonly offset;
+
+  constructor(offset: number) {
+    super();
+    this.offset = offset;
+  }
+}
```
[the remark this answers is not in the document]
zmalatrax: Used in the bitwise builtin, to check that the input cells are Felts

src/errors/memory.ts
```diff
@@ -1,34 +1,29 @@
-import { Relocatable } from 'primitives/relocatable';
 import { SegmentValue } from 'primitives/segmentValue';
 
 export class MemoryError extends Error {}
 
 /** Read a different value (`newValue`) than the already constrained value (`oldValue`) at `address` */
 export class InconsistentMemory extends MemoryError {
-  public readonly address: Relocatable;
+  public readonly offset: number;
```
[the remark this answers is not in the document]
zmalatrax: Artifact from moving it to the class `Segment` and then replace it in `Memory` with the use of Proxy, fixed

src/runners/cairoRunner.ts
```diff
@@ -35,9 +36,12 @@ export class CairoRunner {
     this.programBase = this.vm.memory.addSegment();
     this.executionBase = this.vm.memory.addSegment();
 
+    const builtin_stack = program.builtins
+      .map((builtinName) => BUILTIN_HANDLER[builtinName])
```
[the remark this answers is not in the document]
zmalatrax: Not directly, `map` takes a function as parameter

We can defined a function though:

```typescript
// builtin.ts
const BUILTIN_HANDLER = {};
export const getBuiltin = (name: string) => BUILTIN_HANDLER[name];`

// cairoRunner.ts 
const builtin_stack = program.builtins.map(getBuiltin)...
```

cairo_programs/cairo_0/bitwise_builtin_test.cairo
```diff
@@ -0,0 +1,20 @@
...
+    let (xor_a) = bitwise_xor(12, 10);
+    assert xor_a = 6;
+    let (or_a) = bitwise_or(12, 10);
+    assert or_a = 14;
+
+    let (and_b, xor_b, or_b) = bitwise_operations(9, 3);
+    assert and_b = 1;
+    assert xor_b = 10;
+    assert or_b = 11;
+    return ();
```
ClementWalter, line 19:
```
assert bitwise_ptr.x = x;
assert bitwise_ptr.y = y;
tempvar and = bitwise_ptr.and
// ...
```

ClementWalter (approved):

PR #76, opened 2024-06-04
feat: add ECDSA builtin

zmalatrax:
Solves #67 

- ECDSA Cairo program only works with an hint - cannot do integration test without hint runner
- Currently uses the @noble/curves FpSqrt function, might switch to using @noble/curves more
- Signatures are implemented as a new property of Array (specific to the `ecdsaHandler`), feels a bit hacky but it allows to still be transparent in `memory.ts`
- When working with the `signatures` dict property, one needs to cast the segment `as EcdsaSegment` to avoid TypeScript showing errors `Property 'signatures' does not exist on type 'SegmentValue[]'`


cairo_programs/cairo_0/ecdsa_builtin.cairo
```diff
@@ -0,0 +1,12 @@
+%builtins ecdsa
+
+from starkware.cairo.common.cairo_builtins import SignatureBuiltin
+
+func main{ecdsa_ptr: SignatureBuiltin*}() {
+    %{ ecdsa_builtin.add_signature(ids.ecdsa_ptr.address_, (ids.signature_r, ids.signature_s)) %}
```
ClementWalter, line 6: where does this `signature` come from? (I know we talked on the tact that we currently don't run hint, but here even with hints it wouldn't run)

cairo_programs/cairo_0/ecdsa_builtin.cairo
```diff
@@ -0,0 +1,12 @@
...
+    assert ecdsa_ptr.message = 2718;
+    assert ecdsa_ptr.pub_key = 1735102664668487605176656616876767369909409133946409161569774794110049207117;
```
ClementWalter, line 8: prefer hex string for long numbers

cairo_programs/cairo_0/ecdsa_builtin.cairo
zmalatrax: Indeed, I was inlining the [`verify_ecdsa_signature` function](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/signature.cairo#L6-L19) and forgot about the signature variables

ClementWalter (approved):

PR #77, opened 2024-06-04
feat: add keccak builtin

### LINKED ISSUE #69: feat: implement Keccak builtin

zmalatrax:
Closes #69

### DIFF
--- a/src/builtins/keccak.ts
+++ b/src/builtins/keccak.ts
@@ -1,0 +1,84 @@
+import { bytesToNumberLE, numberToBytesLE } from '@noble/curves/abstract/utils';
+import { concatBytes, u32, u8 } from '@noble/hashes/utils';
+import { keccakP } from '@noble/hashes/sha3';
+
+import { ExpectedFelt } from 'errors/virtualMachine';
+
+import { Felt } from 'primitives/felt';
+import { isFelt } from 'primitives/segmentValue';
+import { BuiltinHandler } from './builtin';
+
+const KECCAK_BYTES = 25;
+const KECCAK_BITS = 200n;
+
+/**
+ * Compute the new state of the keccak-f1600 block permutation on 24 rounds
+ *
+ * - Input: State s, 1600 bits on 8 memory cells (200 bits each)
+ * - Output: State s', 1600 bits on 8 memory cells (200 bits each)
+ * with s' = keccak-f1600(s)
+ */
+export const keccakHandler: BuiltinHandler = {
+  get(target, prop) {
+    if (isNaN(Number(prop))) {
+      return Reflect.get(target, prop);
+    }
+
+    const cellsPerKeccak = 16;
+    const inputCellsPerKeccak = 8;
+
+    const offset = Number(prop);
+    const keccakIndex = offset % cellsPerKeccak;
+    if (keccakIndex < inputCellsPerKeccak) {
+      return target[offset];
+    }
+
+    const inputOffset = offset - keccakIndex;
+    const outputOffset = inputOffset + inputCellsPerKeccak;
+
+    const input = concatBytes(
+      ...target.slice(inputOffset, outputOffset).map((value) => {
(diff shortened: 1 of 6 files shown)


src/builtins/keccak.ts
```diff
@@ -0,0 +1,84 @@
...
+      ...target.slice(inputOffset, outputOffset).map((value) => {
+        if (!isFelt(value)) throw new ExpectedFelt();
+        if (value.toBigInt() >> KECCAK_BITS !== 0n) throw new Error();
+        return numberToBytesLE(value.toBigInt(), 25);
+      })
+    );
+
+    const state = u32(input);
+    keccakP(state);
+    const finalState = u8(state);
```
ClementWalter, line 49: is `keccakP` modifying the `const` state?

src/builtins/keccak.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const outputs = Array.from({ length: 8 }, (_, i) => {
+      const size = i * KECCAK_BYTES;
+      return finalState.slice(size, size + KECCAK_BYTES);
+    }).map(bytesToNumberLE);
```
ClementWalter, line 54: I'd go for somethhing more like

```
(new Array(8)).map((i) => finalState.slice(i * k, (i + 1) * k)).map(bytesToNumberLE)
```

the fact that we just split the finalState in chunks is not visible here

src/builtins/keccak.ts
```diff
@@ -0,0 +1,84 @@
...
+      case 5:
+        target[offset] = new Felt(outputs[5]);
+        break;
+      case 6:
+        target[offset] = new Felt(outputs[6]);
+        break;
+      default:
+        target[offset] = new Felt(outputs[7]);
+        break;
+    }
```
ClementWalter, line 81: why not `target[offset] = new Felt(outputs[keccakIndex - inputCellsPerKeccak]);`

src/builtins/keccak.ts
zmalatrax: Yes, `keccakP` modifies `state`. The `const` keyword blocks re-declaration of objects (i.e. cannot do `state = newState`) but object properties can still be mutated.

Even if `state` values are mutated, `const` keyword avoids assignment overwriting, but `let` could be used for clarity (gotta be consistent throughout the codebase)

Quoting [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const)
> The const declaration creates an immutable reference to a value. It does not mean the value it holds is immutable — just that the variable identifier cannot be reassigned. For instance, in the case where the content is an object, this means the object's contents (e.g., its properties) can be altered. You should understand const declarations as "create a variable whose identity remains constant", not "whose value remains constant" — or, "create immutable [bindings](https://developer.mozilla.org/en-US/docs/Glossary/Binding)", not "immutable values".
>
> Many style guides (including [MDN's](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide/Code_style_guide/JavaScript#variable_declarations)) recommend using const over [let](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) whenever a variable is not reassigned in its scope. This makes the intent clear that a variable's type (or value, in the case of a primitive) can never change. Others may prefer let for non-primitives that are mutated.

src/builtins/keccak.ts
zmalatrax: Creating a new array with `new Array(8)` will create an array of length 8 filled with empty items, which means that `.map` will not iterate over them (thus outputs still being an empty array after the `map`)

ClementWalter (dismissed):

src/builtins/keccak.ts
ClementWalter: indeed, should be `(new Array(8)).fill(0)` which is not better then

ClementWalter (approved):

PR #78, opened 2024-06-05
feat: add Output builtin

### LINKED ISSUE #65: feat: implement Output builtin

zmalatrax:
Closes #65 

- The output builtin has been implemented as a segment which only accepts `Felt`
- A method `getOutput()` has been added to the `CairoRunner` class to return the Output builtin segment if the builtin is in use

To find the index of the Output segment, we rely on the fact that builtins are always initialized after the Program and Execution segments (rationale for the `outputIdx + 2`)

Layouts should enforce the builtins order, with `output` always being the first one. therefore we could simply return `runner.vm.memory.segments[2]` in `getOutput`.
However layout/builtin order is not enforced atm so programs which uses builtins in a different order, like `% builtins bitwise output`  can run on our VM (whereas the cairo VM in python throws as the builtins are not ordered as expected).
This can be updated when implementing layouts.

Python implementation of the Output builtin implements a PublicMemoryPage dictionnary which seems to be only used in the starknetOS & bootloader. There is no need to implement such structure for now imo

### DIFF
--- a/cairo_programs/cairo_0/bitwise_output.cairo
+++ b/cairo_programs/cairo_0/bitwise_output.cairo
@@ -1,0 +1,10 @@
+%builtins output bitwise
+from starkware.cairo.common.bitwise import bitwise_and
+from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
+from starkware.cairo.common.serialize import serialize_word
+
+func main{output_ptr: felt*, bitwise_ptr: BitwiseBuiltin*}() {
+    let (result) = bitwise_and(1, 2);
+    serialize_word(result);
+    return ();
+}
--- a/cairo_programs/cairo_0/bitwise_output.cairo
+++ b/cairo_programs/cairo_0/bitwise_output.cairo
@@ -1,10 +1,11 @@
 %builtins output bitwise
+
 from starkware.cairo.common.bitwise import bitwise_and
 from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
-from starkware.cairo.common.serialize import serialize_word
 
 func main{output_ptr: felt*, bitwise_ptr: BitwiseBuiltin*}() {
     let (result) = bitwise_and(1, 2);
-    serialize_word(result);
+    assert [output_ptr] = result;
+    let output_ptr = output_ptr + 1;
     return ();
 }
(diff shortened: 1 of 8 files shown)


cairo_programs/cairo_0/bitwise_output.cairo
```diff
@@ -0,0 +1,10 @@
+%builtins output bitwise
+from starkware.cairo.common.bitwise import bitwise_and
```
ClementWalter, line 2: add a line break

cairo_programs/cairo_0/bitwise_output.cairo
```diff
@@ -0,0 +1,10 @@
...
+from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
+from starkware.cairo.common.serialize import serialize_word
+
+func main{output_ptr: felt*, bitwise_ptr: BitwiseBuiltin*}() {
+    let (result) = bitwise_and(1, 2);
+    serialize_word(result);
```
ClementWalter, line 8: this function is a scam! (I mean that it's pointless)

```
// Appends a single word to the output pointer, and returns the pointer to the next output cell.
func serialize_word{output_ptr: felt*}(word) {
    assert [output_ptr] = word;
    let output_ptr = output_ptr + 1;
    return ();
}
```

cairo_programs/cairo_0/jmp.cairo
```diff
@@ -0,0 +1,16 @@
...
+func main{output_ptr: felt*}() {
+    jmp test;
+
+    [ap] = 1, ap++;
+    jmp rel 6;
+
+    test:
+    [ap] = 2, ap++;
+
+    serialize_word([ap - 1]);
```
ClementWalter, line 14: see above

ClementWalter (changes requested):
minor changes

cairo_programs/cairo_0/bitwise_output.cairo
ClementWalter, line 1: fyi in the doc https://docs.cairo-lang.org/how_cairo_works/program_input_and_output.html#id2

the output pointer is supposed to be given as arg and not implicit arg

ClementWalter (approved):

PR #85, opened 2024-06-10
feat: add basic cli

### LINKED ISSUE #84: feat: add basic CLI
Add a basic CLI skeleton which allows to:
- Run a Cairo program given its compiled `.json`
- Print the memory
- Print the relocated memory
- Export the relocated memory
- Export the trace

As seen in #60, the relocated memory starts at address 1 for an implementation-specific reason. The `RunOptions` config defaults `relocateOffset` to 0, it should be the same for the CLI. However, for ease of use and expectations from the user (behaviour in other VMs), it might be better to default the `relocateOffset` to 1 in the CLI and add a flag to modify its value

zmalatrax:
Closes #84 

Implements a basic CLI to run the cairo vmin TypeSscript

Dependencies:
- [`commander.js`](https://github.com/tj/commander.js) to build the CLI
- [`consola`](https://github.com/unjs/consola) for an elegant console wrapper

Design choices:
Separate the proof mode and the execution mode by two subcommands rather than a flag
- `cairo-vm-ts run <program.json>` in execution mode (default command so directly using `cairo-vm-ts` will trigger the `run` subcommand)
- `cairo-vm-ts prove <program.json>` in proof mode (TBD)


zmalatrax:
Detailed list of the implemented flags in this [comment](https://github.com/kkrt-labs/cairo-vm-ts/issues/83#issuecomment-2160306432)

README.md
```diff
@@ -92,13 +93,39 @@ performance and safety. While the ones of our TypeScript implementation is
...
+
+### CLI
+
+You can install the CLI `cairo-vm-ts` by doing the following:
+
+1. Clone this repo: `git clone git@github.com:kkrt-labs/cairo-vm-ts.git`
+2. Go to the cloned directory: `cd cairo-vm-ts`
+3. Install the dependencies: `bun install`
+4. Register the package as a _linkable_ package: `bun link`
+5. Link the package: `bun link cairo-vm-ts`
```
ClementWalter, line 106: link two times

README.md
```diff
@@ -92,13 +93,39 @@ performance and safety. While the ones of our TypeScript implementation is
...
+
+Example usage:
+
+```bash
+cairo-vm-ts run fibonacci.json --export-memory fib_mem.bin --print-memory --print-output
```
ClementWalter, line 111: should we keep only `cairo` ? like `cairo run ...`

bin/cli.ts
```diff
@@ -0,0 +1,143 @@
...
+  .option(
+    '--export-memory <MEMORY_FILENAME>',
+    'export the relocated memory, little-endian encoded'
+  )
+  .option('--print-trace', 'print the trace')
+  .option('--print-memory', 'print the non-relocated memory')
+  .option('--print-relocated-memory', 'print the relocated memory')
+  .option('--print-output', 'print the output segment')
+  .action(async (path, options) => {
+    if (!path.match(/\.json$/)) throw new Error('Provided file is not a JSON');
```
ClementWalter, line 47: I think that this should be done in a file parser and not in the body

bin/cli.ts
```diff
@@ -0,0 +1,143 @@
...
+      process.exit(1);
+    }
+
+    consola.info(`Cairo VM TS ${VERSION_CLI} - Execution Mode`);
+
+    const program = parseProgram(fs.readFileSync(path, 'utf-8'));
+    const runner = new CairoRunner(program);
+    const config: RunOptions = {
+      relocate: relocate,
+      relocateOffset: offset,
```
ClementWalter, line 77: can't we keep the same naming ?

bin/cli.ts
```diff
@@ -0,0 +1,143 @@
...
+consola.options = {
+  ...consola.options,
+  formatOptions: {
+    date: false,
+  },
+};
+
+const VERSION_CLI = '0.1.0';
+
+const program = new Command().name('cairo-vm-ts').version(VERSION_CLI);
```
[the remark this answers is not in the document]
zmalatrax: The value in `name` will be the one displayed in the help section 'Usage', so it must match the key in `bin` object from `package.json`

It currently is `cairo-vm-ts` but can be renamed to `cairo` for example

bin/cli.ts
zmalatrax: Sure, but we'll need to either rename the flag `--offset` to `--relocate-offset` with is kinda long imo, or rename the `relocateOffset` property to `offset`. I think renaming `relocateOffset` is ok

bin/cli.ts
zmalatrax: moved to an `argParser` on the argument declaration

package.json
```diff
@@ -2,18 +2,29 @@
...
   "author": "Clément Walter <clement@kakarot.org>",
   "license": "Apache-2.0",
-  "devDependencies": {
-    "bun-types": "^1.1.12",
-    "typescript": "^5.2.2"
+  "main": "dist/index.js",
+  "types": "dist/types/index.d.ts",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "echo \"Error: no entrypoint yet\" && exit 1",
```
ClementWalter, line 12:
```suggestion
```

src/cli.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    consola.info(`Cairo VM TS ${VERSION_CLI} - Execution Mode`);
+
+    const program = parseProgram(fs.readFileSync(String(path), 'utf-8'));
+    const runner = new CairoRunner(program);
+    const config: RunOptions = {
+      relocate: relocate,
+      offset: offset,
+    };
```
ClementWalter, line 87:
```suggestion
    const config: RunOptions = { relocate, offset };
```

ClementWalter (approved):

PR #87, opened 2024-06-13
feat: streamline diff testing

### LINKED ISSUE #86: test: add more Cairo programs
Now that all builtins have been implemented, we should make the test suite more extensive.
The hints haven't been implemented yet, so we limit this to programs which doesn't require hints

- Make a list of all programs without hints from the lambdaclass/cairo-vm cairo programs suite and add them (some imported functions might be inlined (e.g. `serialize_word`...)

zmalatrax:
Closes #86, #88

- Adds all programs from [cairo-vm](https://github.com/lambdaclass/cairo-vm) which have no hints.
- Add a makefile target `compare-memory` which compares the memory files of the different VMs (needs to provide the path to Rust vm CLI as env variable `CAIRO_VM_BIN`, behaviour which might change)
  - This script takes encoded memory as arguments and reconstruct the relocated memory dictionnary for each and compare them.
  Indeed, byte to byte comparison/hash validation is not possible rn as the python order of chunks differs when extra segments are involved.


cairo_programs/cairo_0/big_struct.cairo
ClementWalter, line 1: this doesn't do anything

cairo_programs/cairo_0/call_function_assign_param_by_name.cairo
ClementWalter, line 1: I don't understand the meaning of these test. At the end we run CASM and not cairo and naming has no impact

ClementWalter (changes requested):
- remove failing test
- use all_cairo layout for the rust vm
- use git submodules for the go, rust and zig vms

ClementWalter (approved):

PR #96, opened 2024-06-25
feat: implement hint processor

zmalatrax:
To implement hints, a few things are needed:

- A way to read variable used in hints (`ids.name`), managed by the `IdsManager` class.
This class relies on the `References` parsed from the `ReferenceManager` and the `Identifiers`, given in the compilation artifacts.

- Some hints use constants (or aliases to constants), fetched from the compilation artifact by the `extractConstants` method. (e.g. BigStruct using SquashedDict that are not implemented yet)

When instantiating the CairoRunner, the constants are extracted and the hints are transformed into an exploitable format to be stored as VirtualMachine attributes. 

 Some hints are using data from other hints, the ScopeManager is a stack of dictionaries (scopes) which allows sharing data across hints.

There still needs to add JSDoc and a few util methods might be implemented later on if a group of hints needs it.
At the moment it works but it might not be following our coding guidelines, parts might/should be refactored

### DIFF
--- a/src/hints/hint.ts
+++ b/src/hints/hint.ts
@@ -20,7 +20,7 @@ export class HintProcessor {
     this.scopeManager = new ScopeManager();
   }
 
-  compile(hint: Hint, refManager: ReferenceManager): HintData {
+  static compile(hint: Hint, refManager: ReferenceManager): HintData {
     const references = new Map<string, HintReference>();
     Object.entries(hint.flow_tracking_data.reference_ids).map(
       ([name, index]) => {
--- a/src/hints/hint.ts
+++ b/src/hints/hint.ts
@@ -20,6 +20,12 @@ export class HintProcessor {
     this.scopeManager = new ScopeManager();
   }
 
+  /** Compile an hint from the compilation artifact
+   * to a format that can be easily executed.
+   *
+   * @param hint The hint to be compiled, from the compilation
+   * @param refManager All the program references
+   */
   static compile(hint: Hint, refManager: ReferenceManager): HintData {
     const references = new Map<string, HintReference>();
     Object.entries(hint.flow_tracking_data.reference_ids).map(
@@ -40,6 +46,18 @@ export class HintProcessor {
     };
   }
 
+  /**
+   * Execute an hint based on its code
+   * `hint.code` allows mapping the Python hint code
+   * to its associated function in TypeScript
+   *
+   * `hint.ids` and `vm` provide the context
+   * on which the hint must be executed.
+   *
+   * The virtual machine stores the constants
+   * used by the program.
+   *
+   */
   execute(hint: HintData, vm: VirtualMachine) {
     switch (hint.code) {
       case hintCode.ADD_ECDSA_SIGNATURE:
--- a/src/hints/hint.ts
+++ b/src/hints/hint.ts
@@ -1,10 +1,12 @@
 import { UnknownHint, UnreachableReference } from 'errors/hint';
 
-import { Hint, ProgramConstants, ReferenceManager } from 'vm/program';
+import { Hint, ReferenceManager } from 'vm/program';
+import { VirtualMachine } from 'vm/virtualMachine';
 import { HintReference } from './hintReference';
 import { IdsManager } from './idsManager';
 import { ScopeManager } from './scopeManager';
-import { VirtualMachine } from 'vm/virtualMachine';
+import * as hintCode from './hintCode';
+import { addECDSASignature } from './signature/ecdsa';
 
 export type HintData = {
   ids: IdsManager;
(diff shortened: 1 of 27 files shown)


src/hints/scopeManager.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  /** Set the variable `name` to `value` in the latest scope */
+  set(name: string, value: any) {
+    this.data[this.data.length - 1][name] = value;
+  }
+
+  /** Delete the variable `name` from the latest scope */
+  delete(name: string) {
+    const scope = this.data[this.data.length - 1];
+    delete scope[name];
```
ClementWalter, line 50:
```suggestion
    delete this.data[this.data.length - 1][name];
```

src/vm/program.ts
```diff
@@ -34,27 +35,99 @@ const Identifier = z.object({
...
+export function extractConstants(program: Program): ProgramConstants {
+  const constants: ProgramConstants = new ProgramConstants();
+  Object.entries(program.identifiers).map(([name, identifier]) => {
+    switch (identifier.type) {
+      case 'const':
+        const value = identifier.value;
+        if (!value) throw new ExpectedFelt(value);
+        constants.set(name, value);
+        break;
+      case 'alias':
```
ClementWalter, line 101: could the identifier type by an actual type?

src/vm/program.ts
```diff
@@ -34,27 +35,99 @@ const Identifier = z.object({
...
   return Program.parse(JSON.parse(program));
 }
+
+/**
+ *  Parse the program identifiers
+ * to return a map of constant name to their value
+ */
+export function extractConstants(program: Program): ProgramConstants {
+  const constants: ProgramConstants = new ProgramConstants();
+  Object.entries(program.identifiers).map(([name, identifier]) => {
```
ClementWalter, line 94: don't do swith here, it's the same as the one in `findConstFromAlias`, so just map the resolving function, and `constants.set(...)`

src/hints/hint.ts
```diff
@@ -0,0 +1,52 @@
...
+        const refManagerLen = refManager.references.length;
+        if (index >= refManagerLen)
+          throw new UnreachableReference(index, refManagerLen);
+        const splitName = name.split('.');
+        references.set(
+          splitName[splitName.length - 1],
+          HintReference.parseReference(refManager.references[index])
+        );
+      }
+    );
```
ClementWalter, line 36: refactor to be more functional

src/vm/program.ts
zmalatrax: I don't think so, it is semantic
Here is a list of identifier.type I've seen in compilation artifacts (might be non-exhaustive):
- function
- reference
- type_definition
- struct
- label
- const
- alias
- reference

src/hints/hint.ts
ClementWalter: like

```
const references = Object.entries(hint.flow_tracking_data.reference_ids)
  .map(
    ([name, index]) => {
      const reference = refManager.references[index]
      if (reference === undefined)
        throw new UnreachableReference(index, refManagerLen.references.length);
      return [
        name.split('.').pop(),
        HintReference.parseReference(reference)
      ]
    }
  )
  .reduce((acc, [name, reference]) => {
    acc[name] = reference;
    return acc;
  }, {})
```

src/hints/hintReference.test.ts
```diff
@@ -0,0 +1,369 @@
...
+    );
+  });
+
+  test('should parse cast([ap + (-1)] + [fp + (-3)], felt)', () => {
+    const reference: Reference = {
+      ap_tracking_data: AP_TRACKING_DATA_DEFAULT,
+      pc: 0,
+      value: 'cast([ap + (-1)] + [fp + (-3)], felt)',
+    };
+    expect(HintReference.parseReference(reference)).toEqual(
```
ClementWalter, line 290: parametrize all the tests to `parseReference`

src/hints/hint.ts
ClementWalter, line 1: add docstrings

src/hints/idsManager.ts
```diff
@@ -0,0 +1,151 @@
...
+  constructor(
+    public references: Map<string, HintReference>,
+    public hintApTrackingData: ApTrackingData,
+    public accessibleScopes: string[] = []
+  ) {
+    this.references = references;
+    this.hintApTrackingData = hintApTrackingData;
+  }
+
+  get(name: string, vm: VirtualMachine) {
```
ClementWalter, line 31: looks like feature envy

zmalatrax:
This PR aimed at implementing the Cairo Zero hints with their current format, which forces us to "compile" the hints (parse of references, apply ap tracking and a few other things) that we believe should not be part of the VM.

Therefore, until we have a way to pre-process the current compilation artifacts and provide artifacts which abstract this hint compilation from the VM, I'm closing this PR

PR #97, opened 2024-06-28
feat: Cairo 1 programs and Hint processor

zmalatrax:
WIP
This PR allows executing Cairo programs (v2.6.4, maybe previous version before Sierra changes).
It uses the `*.casm.json` compilation artifacts from the command [`sierra-compile-json`](https://github.com/starkware-libs/cairo/pull/5912)

Currently to run Cairo programs, a flag `--cairo-1` has been added to the `cairo run` subcommand. It successfully ran `add.cairo`.

```bash
cairo run --cairo-1 cairo_programs/cairo/add.casm.json
```

Some refactoring are still needed I think, and then implementing the Cairo 1 hint processor is the next step (which is pretty straightforward).

Still needs to update the CI and add more tests for Cairo programs


zmalatrax:
The trunk CLI tries to format Cairo programs with the Cairo Zero cairo-format

The new cairo format should be configured and used on Cairo programs.

.gitignore
```diff
@@ -4,3 +4,4 @@ dist
 
 # Cairo compilation artifacts
 cairo_programs/**/*.json
+!*.casm.json
```
ClementWalter, line 7: why would we commit cairo artifacts and not cairozero ones?

src/cli.ts
```diff
@@ -30,6 +30,7 @@ program
     })
   )
   .option('-s, --silent', 'silent all logs')
+  .option('--cairo-1', 'Run a Cairo program')
```
ClementWalter, line 33: cairo1 doesn't exist anymore, it's cairo and cairozero/cairo0

why would we need this?

src/errors/hints.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export class InvalidCellRefRegister extends HintError {
+  constructor(cell: CellRef) {
+    super(`Invalid register, expected AP or FP, got ${cell}`);
+  }
+}
+
+export class InvalidOperation extends HintError {
+  constructor(op: string) {
+    super(`Invalid BinOp operator - Expected Add or Mul, received ${op}`);
```
ClementWalter, line 13: what is BinOp ?

src/hints/hintSchema.ts
```diff
@@ -0,0 +1,201 @@
+import { z } from 'zod';
+
+import { Register } from 'vm/instruction';
+import { Felt } from 'primitives/felt';
+
+/**
+ * Types to distinguish ResOp pattern
```
ClementWalter, line 7: what is ResOp ?

src/hints/hintSchema.ts
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * A BinOp is an addition or a multiplication
+ * between two values.
+ * - `a` is a cellRef which should points to a Felt
+ * - `b` is either a Deref or an Immediate, should be a Felt in both cases.
+ */
+export type BinOp = z.infer<typeof binOp>;
+
+/** A ResOp is either a Deref, DoubleDeref, Immediate or BinOp */
+export type ResOp = z.infer<typeof resOp>;
```
ClementWalter, line 132: not sure about this naming, where does this come from ?

src/vm/program.ts
```diff
@@ -58,3 +59,22 @@ export type Program = z.infer<typeof Program>;
 export function parseProgram(program: string): Program {
   return Program.parse(JSON.parse(program));
 }
+
+const Cairo1Program = z.object({
```
ClementWalter, line 63: no "Cairo1"

src/vm/program.ts
```diff
@@ -58,3 +59,22 @@ export type Program = z.infer<typeof Program>;
...
+    .array(z.string())
+    .transform((value) => value.map((v) => new Felt(BigInt(v)))),
+  hints,
+  entrypoint: z.number(),
+  builtins: z.array(z.string()),
+});
+
+type Program2 = z.infer<typeof Cairo1Program>;
+
+export type Cairo1Program = Program2 & { builtins: string[] };
```
ClementWalter, line 76: why this type extension?

src/vm/program.ts
```diff
@@ -58,3 +59,22 @@ export type Program = z.infer<typeof Program>;
...
+  compiler_version: z.string(),
+  bytecode: z
+    .array(z.string())
+    .transform((value) => value.map((v) => new Felt(BigInt(v)))),
+  hints,
+  entrypoint: z.number(),
+  builtins: z.array(z.string()),
+});
+
+type Program2 = z.infer<typeof Cairo1Program>;
```
ClementWalter, line 74: I think that there is one Program type, and that possibly Cairo0 program is Program & { ... }

src/vm/virtualMachine.ts
```diff
@@ -66,10 +68,14 @@ export class VirtualMachine {
...
   /**
    * Execute one step:
+   * - Execute hints at PC
    * - Decode the instruction at PC
    * - Run the instruction
    */
-  step(): void {
+  step(hints?: Hint[]): void {
+    if (hints) {
+      hints.map((hint) => HintProcessor.execute(this, hint));
```
ClementWalter, line 77: I really don't like when `this` is given as first argument of a function, why not then updating the class def of this to have the function as a class method?

src/hints/hintSchema.ts
```diff
@@ -0,0 +1,201 @@
...
+export const testLessThan = z
+  .object({
+    TestLessThan: z.object({ lhs: resOp, rhs: resOp, dst: cellRef }),
+  })
+  .transform(({ TestLessThan: { lhs, rhs, dst } }) => ({
+    type: HintName.TestLessThan,
+    lhs,
+    rhs,
+    dst,
+  }));
```
ClementWalter, line 158: hints should be in their own file with their own type def, and only imported here

src/hints/hintProcessor.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  /**
+   * TestLessThan hint
+   *
+   * Check whether the value at `lhs` is strictly less than the value at `rhs`
+   *
+   * Store the boolean result (0 or 1) at `dst`
+   */
+  static testLessThan(
```
ClementWalter, line 69: move hint actual implementation each in their own file

src/hints/hintSchema.ts
```diff
@@ -0,0 +1,201 @@
...
+export const hintsGroup = z.tuple([z.number(), z.array(hint)]);
+
+/** Zod object to parse an array of grouped hints */
+export const hints = z
+  .array(hintsGroup)
+  .transform((hints) => new Map<number, Hint[]>(hints));
+
+/** Union of all the implemented hints */
+export type Hint = z.infer<typeof hint>;
+
```
ClementWalter, line 188: I do know that a single instruction has an array of hintS and not a single hint. here you mixes hint and hints. `Hint` is one single hint, `Hints = Hint[]` if you want, and a program as hints `Record<Number, Hint[]>`

src/hints/hintSchema.ts
zmalatrax: Taken from the [Cairo compiler](https://github.com/starkware-libs/cairo/blob/a767ed09cf460ba70c689abc2cfd18c0a581fc6c/crates/cairo-lang-casm/src/operand.rs#L31-L48) itself, this is the naming used in the hints.

ResOp was `ResOperand` though, I've shorten it

src/vm/program.ts
zmalatrax: Obsolete code, should be removed

src/vm/virtualMachine.ts
zmalatrax: Will refactor HintProcessor into the VirtualMachine class

src/cli.ts
zmalatrax: Currently the CairoRunner for Cairo Zero and Cairo programs uses different methods (either `fromProgram` or `fromCairo1Program` (should be renamed `fromCairoZeroProgram` and `fromCairoProgram` if any distinction is kept...))

I needed a way to distinguish both programs from the CLI, hence this flag.
Best would be to remove it, otherwise at least rename it `--cairo`. Or the other way around, have a `--cairo-zero` flag which expects a Cairo Zero program and make Cairo programs the default

src/hints/hintSchema.ts
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * `0x1000`
+ */
+export type Immediate = z.infer<typeof immediate>;
+
+/**
+ * A BinOp is an addition or a multiplication
+ * between two values.
+ * - `a` is a cellRef which should points to a Felt
+ * - `b` is either a Deref or an Immediate, should be a Felt in both cases.
```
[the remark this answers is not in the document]
zmalatrax: Yes, except that the bullet point description does not account for the cases when `b` in an immediate, to be exhaustive it gives:
- BinOp (Add): `[register1 + offset1] + [register2 + offset2]` or `[register1 + offset1] + immediate`
- BinOp (Mul): `[register1 + offset1] * [register2 + offset2]` or `[register1 + offset1] * immediate`

Using the variables `a` and `b` instead, it would be equivalent to:
- BinOp(Add): `[a] + b`
- BinOp(Mul): `[a] * b`

src/vm/program.ts
zmalatrax: We could make a base Zod object with all shared fields between Cairo Zero and Cairo compilation artifacts and then extend it for both specific fields. I'll propose a refactor

src/hints/hintSchema.ts
zmalatrax: Not sure to understand the issue, it is about the naming of `hints` as `z.array(hintsGroup)` rather than something else and have `hints` as `z.array(hints)` ?

(The field "hints" of a program artifacts is an array of tuple of hints: i.e. `Array<Record<Numer, Hint[]>>`

src/cli.ts
```diff
@@ -30,6 +30,10 @@ program
     })
   )
   .option('-s, --silent', 'silent all logs')
+  .option('--cairo', 'Run a Cairo program')
```
ClementWalter, line 33: I'm still wondering why this is required. I think that we should have one `Program` type and that at the VM level, there is no Cairo/CairoZero but CASM. If for now we can't (for some reasons to be better explained) have this, at lease the parser of the input should be able to figure this out

src/hints/allocSegment.ts
```diff
@@ -0,0 +1,28 @@
+import { z } from 'zod';
+
+import { VirtualMachine } from 'vm/virtualMachine';
+import { cellRef, CellRef } from 'hints/hintParamsSchema';
+import { HintName } from 'hints/hintName';
+
+/** Zod object to parse AllocSegment hint */
+export const allocSegmentZod = z
```
ClementWalter, line 8:
```suggestion
export const allocSegmentParser = z
```

suggesting to remove the `Zod` name as we may change one day the parser lib, I prefer to keep the naming explicit about the function that the lib

src/runners/cairoRunner.ts
```diff
@@ -112,7 +162,7 @@ export class CairoRunner {
   }
 
   getOutput() {
-    const builtins = this.program.builtins;
+    const builtins = (this.program as CairoZeroProgram).builtins;
```
ClementWalter, line 165: why is this cast needed?

src/hints/allocSegment.ts
zmalatrax: Indeed, will make the changes in the implemented hints (AllocSegment & TestLessThan)

src/runners/cairoRunner.ts
zmalatrax: Cairo Zero artifacts have a global `builtins` field containing the 'output' builtin, whereas Cairo/Starknet artifacts have a `builtins` field nested in each function entrypoint which doesn't contain the 'output' builtin.

The output builtin is always used in `proof_mode` and for Starknet contracts. In the `cairo-vm` , for Cairo programs in ExecutionMode, it requires a flag (see `cairo1-run`), only inputs and return values of the type `Array<Felt252>` are expected/accepted.

The `getOutput()` method only has the logic for Cairo Zero artifacts. We could differenciate both artifacts and throw an exception for CairoProgram until the logic around input/return types is implemented.

The cast is needed to avoid a TypeScript error because there isn't a global `builtins` field in both artifacts.

CHANGELOG.md of the `cairo-vm`:
> [1.0.0-rc4] - 2024-07-05
>  * feat(BREAKING): Serialize inputs into output segment in cairo1-run crate:
>    * Checks that only `Array<Felt252>` can be received by the program main function when running with with either `--proof_mode` or `--append_return_values`.
>    * Copies the input value to the output segment right after the output in the format `[array_len, arr[0], arr[1],.., arr[n]]`.
>  * feat(BREAKING): Serialize `Array<Felt252>` return value into output segment in cairo1-run crate:
>    * Checks that only `PanicResult<Array<Felt252>>` or `Array<Felt252>` can be returned by the program when running with either `--proof_mode` or `--append_return_values`.
>  * Serializes return values into the output segment under the previous conditions following the format:
>    * `PanicResult<Array<Felt252>>` -> `[panic_flag, array_len, arr[0], arr[1],.., arr[n]]`
>    * `<Array<Felt252>` -> `[array_len, arr[0], arr[1],.., arr[n]]`
> 
> [1.0.0-rc2] - 2024-05-02
> * feat: Add flag to append return values to output segment when not running in proof_mode [#1646](https://github.com/lambdaclass/cairo-vm/pull/1646)
>  * Adds the flag `append_return_values` to both the CLI and `Cairo1RunConfig` struct.
>  * Enabling flag will add the output builtin and the necessary instructions to append the return values to the output builtin's memory segment.

src/hints/testLessThan.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        register: Register.Ap,
+        offset: -1,
+      },
+    });
+  });
+
+  test.each([
+    [new Felt(-2n), new Felt(0n)],
+    [new Felt(2n), new Felt(1n)],
+  ])('should properly execute TestLessThan hint', (lhs, expected) => {
```
ClementWalter, line 55:
```suggestion
  test.each([
    [new Felt(BigInt('0x100000000')), new Felt(1n)],
    [new Felt(BigInt('0x100000001')), new Felt(0n)],
    [new Felt(-2n), new Felt(0n)],
    [new Felt(2n), new Felt(1n)],
  ])('should properly execute TestLessThan hint', (lhs, expected) => {
```

src/runners/cairoRunner.ts
```diff
@@ -29,16 +36,20 @@ export class CairoRunner {
     offset: 0,
   };
 
-  constructor(program: Program) {
+  constructor(
+    program: Program,
+    bytecode: Felt[],
+    mainOffset: number = 0,
```
ClementWalter, line 42:
```suggestion
    initialPc: number = 0,
```

src/hints/testLessThan.test.ts
zmalatrax: It is a strict inequality so `'0x100000000'` should return False (`0n`), but having cases around the bound should definitely be added. I'll add 0x100000000 and 0x0ffffffff

ClementWalter (approved):

PR #106, opened 2024-07-17
feat: implement dict hints primitives

### LINKED ISSUE #102: AllocFelt252Dict

### LINKED ISSUE #103: feat: implement SegmentArena builtin

### LINKED ISSUE #104: feat: implement dictionnary manager/tracker

### LINKED ISSUE #105: feat: implement ScopeManager (ExecutionScopes)

### LINKED ISSUE #107: Felt252DictEntryInit

### LINKED ISSUE #108: Felt252DictEntryUpdate

### LINKED ISSUE #109: GetSegmentArenaIndex

### LINKED ISSUE #110: InitSquashData

### LINKED ISSUE #111: GetCurrentAccessIndex

### LINKED ISSUE #112: ShouldSkipSquashLoop

### LINKED ISSUE #113: GetCurrentAccessDelta

### LINKED ISSUE #114: ShouldContinueSquashLoop

### LINKED ISSUE #115: GetNextDictKey

### LINKED ISSUE #116: AssertLeFindSmallArcs

### LINKED ISSUE #117: AssertLeIsFirstArcExcluded

### LINKED ISSUE #118: AssertLeIsSecondArcExcluded

zmalatrax:
Closes #102, closes #103, closes #104, closes #105, closes #107, closes #108, closes #109, closes #110, closes #111, closes #112, closes #113, closes #114, closes #115, closes #116, closes #117, closes #118

To execute and implement hints related to dictionnaries, a few things must be implemented first:

- `ScopeManager`, to access variables across different hints (used in `AssertLeFindSmallArc`, `AssertLeIsFirstArcExcluded` and `AssertLeIsSecondArcExcluded`).
- `DictManager` a map of dictionnaries, where allocated dictionnaries are stored during execution
- `SquashedDictManager`, handles the squashing of the dictionnaries, needed to end the run and generate correct traces.

Once done, dictionnaries can be handled.

| Implemented hint | Usage |
| --- | --- |
`AllocFelt252Dict` | Dictionnary initialization
`Felt252DictEntryInit` | Dictionnary usage (methods from `dict.cairo`)
`Felt252DictEntryUpdate` | Dictionnary usage (methods from `dict.cairo`)
`GetSegmentArenaIndex` | Dictionnary finalization
`InitSquashData` | Dictionnary finalization
`GetCurrentAccessIndex` | Dictionnary finalization
`ShouldSkipSquashLoop` | Dictionnary finalization
`GetCurrentAccessDelta` | Dictionnary finalization
`ShouldContinueSquashLoop` | Dictionnary finalization
`GetNextDictKey` | Dictionnary finalization
`AssertLeFindSmallArcs` | Dictionnary finalization
`AssertLeIsFirstArcExcluded` | Dictionnary finalization
`AssertLeIsSecondArcExcluded` | Dictionnary finalization


src/hints/squashedDictManager.ts
```diff
@@ -0,0 +1,70 @@
...
+  /** Add `index` to the indices taken by `key` */
+  insert(key: Felt, index: Felt) {
+    const keyStr = key.toString();
+    const indices = this.keyToIndices.get(keyStr);
+    if (!indices) {
+      this.keyToIndices.set(keyStr, [index]);
+    } else {
+      indices.push(index);
+    }
+  }
```
ClementWalter, line 32: maybe we can consider using a `DefaultMap` as per examples here https://stackoverflow.com/questions/51319147/map-default-value instead

src/hints/dict/initSquashData.ts
```diff
@@ -0,0 +1,97 @@
...
+  ptrDiff: ResOperand,
+  nAccesses: ResOperand,
+  bigKeys: CellRef,
+  firstKey: CellRef
+) => {
+  const dictAccessSize = 3;
+  const address = vm.getPointer(...vm.extractBuffer(dictAccesses));
+  if (Number(vm.getResOperandValue(ptrDiff)) % dictAccessSize)
+    throw new Error(
+      'Accessess array size must be divisible by the dict size (3)'
```
enitrat, line 74: typo
```suggestion
      'Accesses array size must be divisible by the dict size (3)'
```

src/hints/dict/initSquashData.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+export const initSquashData = (
+  vm: VirtualMachine,
+  dictAccesses: ResOperand,
+  ptrDiff: ResOperand,
+  nAccesses: ResOperand,
+  bigKeys: CellRef,
+  firstKey: CellRef
+) => {
+  const dictAccessSize = 3;
```
enitrat, line 70: move out of function and import in all relevant files
`const DICT_ACCESS_SIZE = 3`

src/hints/dict/initSquashData.ts
```diff
@@ -0,0 +1,97 @@
...
+ * - Assert the smallest (first) key to `firstKey` address
+ *
+ */
+export const initSquashData = (
+  vm: VirtualMachine,
+  dictAccesses: ResOperand,
+  ptrDiff: ResOperand,
+  nAccesses: ResOperand,
+  bigKeys: CellRef,
+  firstKey: CellRef
```
enitrat, line 68: we are not going to refactor for consistency with the rest of the codebase but I think that ideally the signatures of these functions could've been `export const initSquashData(vm, {dictAccesses, ptrDiff, nAccess, bigKeys, firstKey})` which feels more aligned with the zod schema (initSquashDataInput would be a single object)

src/hints/dict/initSquashData.ts
```diff
@@ -0,0 +1,97 @@
...
+  for (let i = 0; i < nbAccesses; i++) {
+    const key = vm.memory.get(address.add(i * dictAccessSize));
+    if (!key || !isFelt(key)) throw new ExpectedFelt(key);
+    vm.squashedDictManager.insert(key, new Felt(BigInt(i)));
+  }
+  vm.squashedDictManager.keyToIndices.forEach((values, key) => {
+    values.reverse();
+    vm.squashedDictManager.keys.push(new Felt(BigInt(key)));
+  });
+  vm.squashedDictManager.keys.sort((a, b) => (a < b ? 1 : a > b ? -1 : 0));
```
enitrat, line 86: implement `compare` on `Felt`, then:
`vm.squashedDictManager.keys.sort((a, b) => b.compare(a));`

src/hints/dict/initSquashData.ts
```diff
@@ -0,0 +1,97 @@
...
+export type InitSquashData = z.infer<typeof initSquashDataParser>;
+
+/**
+ * Initialize the squashed dictionnary
+ *
+ * - Insert all accessed keys
+ * - Assert the biggest key to `bigKeys` address
+ * - Assert the smallest (first) key to `firstKey` address
+ *
+ */
```
enitrat, line 61: JSDoc should mainly provide information on the expected parameters. As a first read I don't know what `bigKeys` is

src/hints/dict/initSquashData.ts
```diff
@@ -0,0 +1,97 @@
...
+  vm: VirtualMachine,
+  dictAccesses: ResOperand,
+  ptrDiff: ResOperand,
+  nAccesses: ResOperand,
+  bigKeys: CellRef,
+  firstKey: CellRef
+) => {
+  const dictAccessSize = 3;
+  const address = vm.getPointer(...vm.extractBuffer(dictAccesses));
+  if (Number(vm.getResOperandValue(ptrDiff)) % dictAccessSize)
```
enitrat, line 72: nit: overall readability could gain from using descriptive variable names, here e.g. `if (ptrDiffValue % DICT_ACCESS_SIZE)` and separating different parts of the functions with an empty line

src/hints/dict/getCurrentAccessDelta.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * GetCurrentAccessDelta hint
+ *
+ * Assert to memory the difference between the current index to the next one.
+ *
+ */
+export type GetCurrentAccessDelta = z.infer<typeof getCurrentAccessDeltaParser>;
+
+/**
+ * Assert `currIndex - prevIndex - 1` to `indexDeltaMinusOne`
```
enitrat, line 25: jsdoc could explain what `indexDeltaMinusOne` is

src/hints/dict/felt252DictEntryUpdate.ts
```diff
@@ -0,0 +1,55 @@
...
+ * `dict[key] = value`
+ *
+ */
+export const felt252DictEntryUpdate = (
+  vm: VirtualMachine,
+  dictPtr: ResOperand,
+  value: ResOperand
+) => {
+  const address = vm.getPointer(...vm.extractBuffer(dictPtr));
+  const key = vm.memory.get(address.sub(3));
```
enitrat, line 48: `3` should be a named const

src/hints/dict/felt252DictEntryUpdate.ts
```diff
@@ -0,0 +1,55 @@
...
+>;
+
+/**
+ * Get the dictionnary `dict` at `dictPtr`
+ *
+ * Get the key at address `dict - 3`
+ *
+ * `dict[key] = value`
+ *
+ */
```
enitrat, line 41:
```suggestion
/**
 * Updates an existing dictionary entry with a new value.
 * 
 * @param vm - The virtual machine instance.
 * @param dictPtr - Pointer to the dictionary access.
 * @param value - The new value to be set.
 * @throws {ExpectedFelt} If the key is not a valid Felt.
 */
```

This format would be better for all JSDoc

src/hints/dict/felt252DictEntryInit.ts
```diff
@@ -0,0 +1,51 @@
...
+  vm: VirtualMachine,
+  dictPtr: ResOperand,
+  key: ResOperand
+) => {
+  const address = vm.getPointer(...vm.extractBuffer(dictPtr));
+  const keyValue = vm.getResOperandValue(key).toString();
+  const dict = vm.getDict(address);
+  const prevValue = dict.get(keyValue) || new Felt(0n);
+  dict.set(keyValue, prevValue);
+  vm.memory.assertEq(address.add(1), prevValue);
```
enitrat, line 50: `1` should be a named const

enitrat (commented):
Considering that this implementation is the "front-door" to the CairoVM, I would double-down on explicit and well-formed JSDoc for every functions

src/runners/cairoRunner.ts
```diff
@@ -65,9 +65,22 @@ export class CairoRunner {
...
     this.builtins = builtins;
-    const builtin_stack = builtins
-      .map(getBuiltin)
-      .map((builtin) => this.vm.memory.addSegment(builtin));
+    const builtin_stack = builtins.map((builtin) => {
+      const handler = getBuiltin(builtin);
+      if (builtin === 'segment_arena') {
+        const info = [
+          this.vm.memory.addSegment(handler),
+          new Felt(0n),
```
enitrat, line 73: what are these constants?

src/hints/dict/felt252DictEntryInit.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export const felt252DictEntryInit = (
+  vm: VirtualMachine,
+  dictPtr: ResOperand,
+  key: ResOperand
+) => {
+  const address = vm.getPointer(...vm.extractBuffer(dictPtr));
+  const keyValue = vm.getResOperandValue(key).toString();
+  const dict = vm.getDict(address);
+  const prevValue = dict.get(keyValue) || new Felt(0n);
```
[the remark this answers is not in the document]
zmalatrax: Setting the previous value to `0` when initializing a key is the convention for Cairo dicts afaik

Taken from the [Cairo book](https://book.cairo-lang.org/ch03-02-dictionaries.html):
> For each time we interact with a Felt252Dict<T>, a new Entry<T> will be registered:
>
>    A `get` would register an entry where there is no change in state, and previous and new values are stored with the same value.
>    An `insert` would register a new `Entry<T>` where the `new_value` would be the element being inserted, and the `previous_value` the last element inserted before this. In case it is the first entry for a certain key, then the previous value will be zero.

[Rust implementation](https://github.com/lambdaclass/cairo-vm/blob/37ea72977dccbc2b90b8b7534c1edabd2e2fef79/vm/src/hint_processor/cairo_1_hint_processor/hint_processor.rs#L783-L785), [Go implementation](https://github.com/NethermindEth/cairo-vm-go/blob/be3418769457710ad529a5bb59c461b720cce380/pkg/hintrunner/core/hint.go#L826-L830)

src/hints/dict/allocFelt252Dict.ts
```diff
@@ -0,0 +1,48 @@
...
+  const arenaPtr = vm.getPointer(...vm.extractBuffer(segmentArenaPtr));
+  const dictNumber = vm.memory.get(arenaPtr.sub(2));
+  if (!dictNumber || !isFelt(dictNumber)) throw new ExpectedFelt(dictNumber);
+
+  const infoPtr = vm.memory.get(arenaPtr.sub(3));
+  if (!infoPtr || !isRelocatable(infoPtr))
+    throw new ExpectedRelocatable(infoPtr);
+
+  const newDict = vm.newDict();
+  vm.memory.assertEq(infoPtr.add(dictNumber.mul(new Felt(3n))), newDict);
```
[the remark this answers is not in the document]
zmalatrax: Yes, 3n should be refactored into a named constant `DICT_ACCESS_SIZE`

For `dictNumber` and `infoPtr` offsets to compute them, I don't think that making named constants is added-value here imho

It boils down to how the segment arena builtin works by block of 3 cells:
- First cell is the base address of the InfoPtr
- Second cell is the current number of dictionnaries
- Third cell is the current number of squashed dictionnaries

I agree on explaining the memory layout in the jsdoc

src/hints/dict/initSquashData.ts
zmalatrax: mmh I disagree on having the inputs as a single object for hint parameters. The current zod schema flatten the hint name with its parameters, nesting them in another object would only add complexity.

For reference, the InitSquashData hint found in compilation artifacts:
```json
{
    "InitSquashData": {
        "dict_accesses": { "Deref": { "register": "FP", "offset": -4 } },
        "ptr_diff": { "Deref": { "register": "FP", "offset": 0 } },
        "n_accesses": { "Deref": { "register": "AP", "offset": -1 } },
        "big_keys": { "register": "FP", "offset": 2 },
        "first_key": { "register": "FP", "offset": 1 }
    }
}
```

Current zod schema gives
```json
{
    "type": "InitSquashData",
    "dictAccesses": {},
    "ptrDiff": {},
}
```

and your proposal would give:
```json
{
   "type": "InitSquashData",
    "initSquashDataInput": {
        "dictAccesses": {},
        "ptrDiff": {},
    }
}
```
It will reduce readability in the core logic function of the hint, either having to destruct the object in multiple variables `const {dictAccesses, ptrDiff} = initSquashDataInput`,  or having an unnecessary verbose name `initSquashDataInput.dictAccesses`...)

src/runners/cairoRunner.ts
zmalatrax: The segment arena builtin initialization:
- base address of the info segment
- current number of dictionnaries (i.e. 0)
- current number of squashed dictionnaries (i.e. 0)

src/hints/dict/felt252DictEntryInit.ts
zmalatrax: `address` is a pointer to the next dict access: for a dictionnary on segment 8, for the first key initialized, `address == new Relocatable(8, 0)` is (`8:0`) ; next entry `address` will be `8:3` etc

A dictionnary access works by block of 3 cells:
- First cell is the key
- Second cell is the previous value, initialized at 0 if undefined --> `Felt252DictEntryInit` assert this value in memory
- Third cell is the current value --> `Felt252DictEntryUpdate` assert this value in memory

Naming `1` would not make much sense here, I'd rather name `address.add(1)` as `prevValueAddress`.
Though leaving it inlined is better imo

src/hints/dict/initSquashData.ts
zmalatrax: Indeed, makes me realize that it should be a defined error extending `HintError` rather than a plain `Error`.

src/hints/dict/initSquashData.ts
zmalatrax: Agree, as it is used in multiple hints (`InitSquashData` and `AssetLeFindSmallArcs`)

src/hints/dict/initSquashData.ts
enitrat: fair enough :)

src/hints/dict/felt252DictEntryInit.ts
enitrat: key_offset is 0, prev_value_offset is 1, new_value_offset is 2, I still think that something like `vm.memory.assertEq(address.add(PREV_VALUE_OFFSET), prevValue);` would be better as it conveys in a single word the entire explanation given above

zmalatrax: legit, will refactor Dictionary and related constants to their own file `dictionary.ts`

src/hints/dict/felt252DictEntryInit.ts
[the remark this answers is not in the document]
enitrat: this doesn't surprise me for cairo1 i'm a bit more surprised for cairo0 (but perhaps i'm out of topic and this is only for cairo1 stuff? in which case you can disregard my comment)

```
func default_dict_new(default_value: felt) -> (res: DictAccess*) {
    %{
        if '__dict_manager' not in globals():
            from starkware.cairo.common.dict import DictManager
            __dict_manager = DictManager()

        memory[ap] = __dict_manager.new_default_dict(segments, ids.default_value)
    %}
    ap += 1;
    return (res=cast([ap - 1], DictAccess*));
}
```

The dict manager registers what the default value of a default_dict is - and it's not necessarily zero.

see: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/dict.py#L44-L55

```
    def new_default_dict(self, segments, default_value, temp_segment: bool = False):
        """
        Creates a new Cairo default dictionary.
        """
        base = segments.add_temp_segment() if temp_segment else segments.add()
        assert base.segment_index not in self.trackers
        self.trackers[base.segment_index] = DictTracker(
            data=defaultdict(lambda: default_value),
            current_ptr=base,
        )
        return base
```

data is implemented as a defaultdict, where if a value is not initialized in the dict, the lambda function will be executed and will return the default value.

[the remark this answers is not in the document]
enitrat: ok great! thank you

src/runners/cairoRunner.ts
zmalatrax: renamed `info` to `initialValues` which is more meaningful

enitrat (approved):

PR #119, opened 2024-07-25
feat: add layouts

### LINKED ISSUE #89: feat: add Layouts
Implement the different layouts and related logic.

Must document the layout components (what, why, where): attributes, methods where they are used, expected behavior etc...

zmalatrax:
Closes #89 

- Add builtins validation against the layout.
- Add a flag `-l, --layout` in the CLI
- `ratio` values which were living in the BuiltinInstanceDef in other implementations are refactored into the Layout as the `ratios` dictionnary.
- `getBuiltinSegment` method which allows extracting any memory segment which represents a builtin. It needs to know which builtins have been initialized by the run.
- `builtins` attribute for CairoRunner, needed for `getBuiltinSegment`

NOTE: The `builtins` attribute currently stores the builtins used by the called program entrypoint (`program.builtins`). It works well in ExecutionMode as only the builtins in the program are used. However, in Proof Mode, all the builtins from the chosen layout are initialized, that won't be represented by the current `builtins` attribute.
When introducting the proof mode, this `builtins` attribute initialization should be refactored to take this into account.

### DIFF
--- a/src/runners/layout.ts
+++ b/src/runners/layout.ts
@@ -125,6 +125,8 @@ export const layouts: { [key: string]: Layout } = {
   },
 };
 
+export const ALL_LAYOUTS = Object.keys(layouts);
+
 export function isSubsequence(
   subsequence: string[],
   sequence: string[]
--- a/src/runners/layout.ts
+++ b/src/runners/layout.ts
@@ -11,11 +11,13 @@ export type DilutedPool = {
 /**
  * Layout provides a configuration when generating a proof
  * - Builtins: List of available builtins.
- *   segment_arena, gas_builtin and system are not proven.
  * - rcUnits: Range Check Units.
  * - publicMemoryFraction: The ratios total memory cells over public memory cells.
  * - dilutedPool: The diluted pool, used for additionnal checks on Bitwise & Keccak
  * - ratios: Dictionnary mapping each builtin name to its ratio.
+ *
+ * NOTE: The builtins `segment_arena`, `gas_builtin` and `system` which can be found
+ * in the program builtins are not part of the layouts because they're not proven as such.
  *
  * NOTE: A ratio defines the number of steps over the number of builtin instances.
  * For every ratio steps, we have one instance.
--- a/src/runners/layout.ts
+++ b/src/runners/layout.ts
@@ -1,0 +1,144 @@
+export type DilutedPool = {
+  unitsPerStep: number;
+  spacing: number;
+  nBits: number;
+};
+
+export type Layout = {
+  builtins: string[];
+  rcUnits: number;
+  publicMemoryFraction: number;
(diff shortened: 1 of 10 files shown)


src/runners/layout.ts
```diff
@@ -0,0 +1,146 @@
...
+      if (index === -1) {
+        acc.found = false;
+      } else {
+        acc.sequence = acc.sequence.slice(index + 1);
+      }
+      return acc;
+    },
+    { sequence, found: true }
+  ).found;
+}
```
ClementWalter, line 146: this looks a bit complicted to me. What about
so default empty sequence should return `true` as per your code. I propose

```
export function isSubsequence(
  subsequence: string[],
  sequence: string[]
): boolean {
    if (!sequence) return true;

    const index = sequence.find(subsequence[0])
    if (index === -1) return false;

    return sequence.slice(index, index + subsequence.length) === subsequence
}
```

zmalatrax: `!sequence` checks if sequence is undefined, to chech for empty array we have to check the value of `array.length`

It should be the length of `subsequence` rather than `sequence`, as the program builtins represent the subsequence while the sequence represent the layout builtins.

Array methods like `find` and `findIndex` takes a predicate as argument

Not sure to understand what does `slice(index, index + subsequence.length)` do, it should take all the remaining elements of `sequence` after the previously found `index`.
Plus, the `subsequence` might not exactly equal the reamining `sequence`.

But I got the idea, we can do it recursively, I'm proposing:

```typescript
export function isSubsequence(
  subsequence: string[],
  sequence: string[]
): boolean {
  if (!subsequence.length) return true;

  const index = sequence.findIndex((name) => name === subsequence[0]);
  if (index === -1) return false;

  return isSubsequence(subsequence.slice(1), sequence.slice(index + 1));
}
```

src/runners/layout.ts
```diff
@@ -0,0 +1,243 @@
...
+export type DilutedPool = {
+  unitsPerStep: number;
+  spacing: number;
+  nBits: number;
+};
+
+/**
+ * Layout provides a configuration when generating a proof
+ * - Builtins: List of available builtins.
+ *   segment_arena, gas_builtin and system are not proven.
```
enitrat, line 14: in doc but not in type?

enitrat (dismissed):

src/runners/layout.ts
zmalatrax: Those three builtins (`segment_arena`, `gas_builtin` and `system`) are not in the layouts because they're not proven as such (no AIR for them in the prover)

Wanted to make a note on it in docs but it is kinda unclear. I'll refactor it to a NOTE for clarity

enitrat: ah sorry my bad

enitrat (approved):

PR #127, opened 2024-08-07
feat: Add div mod hint

coxmars:
Hi @zmalatrax,

I've made significant progress on the `DivMod` hint implementation, including updates to the `hintName.ts`, `hintSchema.ts`, `hintHandler.ts`. The core logic is nearly complete, but a few final adjustments are still needed in `divMod.ts`. Also, the `verified` tag is working according with the contribution guidelines.

#### Status
- Add div mod hint working as expected.


coxmars:
#### Update 📌

Hi @zmalatrax

- Implemented and tested the `DivMod` hint.
- All related tests for `DivMod` have passed successfully.

- Here are the test results:
  
image

- Please review the changes and let me know if any additional modifications are needed, also can you explain me a little bit the integration test with Cairo I think it is the only I need to complete it 🫡

src/hints/math/divMod.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    const quotientValue = new Felt(lhsValue / rhsValue);
+    const remainderValue = new Felt(lhsValue % rhsValue);
+    
+    const quotientAddr = vm.cellRefToRelocatable(quotientAddress);
+    const remainderAddr = vm.cellRefToRelocatable(remainderAddress);
+
+    vm.memory.assertEq(quotientAddr, quotientValue);
+    vm.memory.assertEq(remainderAddr, remainderValue);
```
zmalatrax, line 64: No need to declare new variables for quotient and remainder addresses, you can directly call `vm.memory.assertEq(vm.cellRefToRelocatable(quotient), quotientValue)`

src/hints/math/divMod.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    const remainderExpected = new Felt(2n);
+
+    const lhsAddr = new Relocatable(0, 0);
+    const rhsAddr = new Relocatable(0, 1);
+    const quotientAddr = new Relocatable(0, 2);
+    const remainderAddr = new Relocatable(0, 3);
+
+    vm.memory.assertEq(lhsAddr, lhsValue);
+    vm.memory.assertEq(rhsAddr, rhsValue);
+    vm.ap = new Relocatable(0, 0);
```
zmalatrax, line 85: Why storing the values in the segment 0 while `ap` points to the segment 1?

Leave `ap` as is (it points to `Relocatable(1,0) by default)

Use `vm.ap.add()` to compute the addresses

```suggestion
vm.memory.assertEq(vm.ap, lhsValue);
vm.memory.assertEq(vm.ap.add(1), rhsValue);
```

src/hints/math/divMod.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    const remainderAddr = new Relocatable(0, 3);
+
+    vm.memory.assertEq(lhsAddr, lhsValue);
+    vm.memory.assertEq(rhsAddr, rhsValue);
+    vm.ap = new Relocatable(0, 0);  
+
+    divMod(vm, hint.leftHandSide, hint.rightHandSide, hint.quotientAddress, hint.remainderAddress);
+
+    expect(vm.memory.get(quotientAddr)).toEqual(quotientExpected);
+    expect(vm.memory.get(remainderAddr)).toEqual(remainderExpected);
```
zmalatrax, line 90: use `vm.cellRefToRelocatable(hint.quotient)` rather than `quotientAddr`, same for the remainder

src/hints/math/divMod.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  test('should properly execute DivMod hint', () => {
+    const hint = divModParser.parse(DIV_MOD_HINT);
+    const vm = new VirtualMachine();
+    vm.memory.addSegment();
+    vm.memory.addSegment();
+
+    const lhsValue = new Felt(17n);
+    const rhsValue = new Felt(5n);
+    const quotientExpected = new Felt(3n);
+    const remainderExpected = new Felt(2n);
```
zmalatrax, line 76: Make a few different cases by parameterizing the test, it can be done with `test.each`, the [TestLessThan test](https://github.com/kkrt-labs/cairo-vm-ts/blob/e133561a28ea1f269f6efd1af8bb657007c20b74/src/hints/math/testLessThan.test.ts#L53-L58) can be used as a reference
And here is the [doc](https://bun.sh/docs/test/writing#test-each)

zmalatrax (commented):
great :)

coxmars:
> @coxmars Looks like your typescript files aren't well formatted, we're using Trunk, I'd recommend you installing their [extension](https://docs.trunk.io/code-quality/ide-integration/vs-code) if you're on VSCode or Neovim
> 
> Otherwise, you can add their [CLI ](https://docs.trunk.io/code-quality/usage) and run `trunk fmt`.

Thanks for the feedback, I will apply the requested changes asap 🫡

zmalatrax:
About the integration test, you need to add a Cairo program `divmod.cairo` in the `cairo_programs/cairo/hints` folder that makes use of the DivMod hint.

It should be a simple program focused on this hint, such as
```cairo
fn main() {
    let _ = 17_u32 / 12_u32;
    let _ = 17_u32 % 12_u32;
}
```

coxmars:
Hi @zmalatrax I'm trying to execute make compile and I got this error:

image

- I already have Bun, Poetry, I ran poetry install for dependencies and I downloaded cairo-lang too and also Python 3.10 so do you have an idea what is the problem?

coxmars:
Hi @zmalatrax I was able to solve some problems running the integration test haha, it took me a while but I got it done, unit test and integration test ready plus the feedback given, as you can see below it is working so I will open this draft PR to "Ready for review" 🫡

image

src/hints/math/divMod.ts
```diff
@@ -0,0 +1,66 @@
...
+  rhs: ResOperand,
+  quotient: CellRef,
+  remainder: CellRef
+) => {
+  const lhsValue = vm.getResOperandValue(lhs).toBigInt();
+  const rhsValue = vm.getResOperandValue(rhs).toBigInt();
+
+  if (rhsValue === 0n) {
+    throw new Error('Division by zero');
+  }
```
zmalatrax, line 59: Use `!rhsValue` instead of `rhsValue === 0n` in the if condition

Use the `CannotDivideByZero()` error defined in `errors/primitives.ts`

Avoid using brackets when it is a simple if condition which throws an error

```suggestion
if (!rhsValue) throw new CannotDivideByZero()`
```

src/hints/math/divMod.test.ts
```diff
@@ -0,0 +1,91 @@
...
+        offset: 2,
+      },
+      remainder: {
+        register: Register.Ap,
+        offset: 3,
+      },
+    });
+  });
+
+  test.each([
```
enitrat, line 65: test cases are not covering edge cases

coxmars:
Hi @zmalatrax I already changed the condition as you said and @enitrat I added a new edge case test, if you need something else to be improved just tell me, thanks 🫡

src/hints/math/divMod.test.ts
```diff
@@ -88,4 +88,21 @@ describe('DivMod', () => {
...
+
+    const lhsValue = new Felt(17n);
+    const rhsValue = new Felt(0n);
+
+    vm.memory.assertEq(vm.ap, lhsValue);
+    vm.memory.assertEq(vm.ap.add(1), rhsValue);
+
+    expect(() => {
+      divMod(vm, hint.lhs, hint.rhs, hint.quotient, hint.remainder);
+    });
```
zmalatrax, line 106: You're expecting the  `divMod` function to throw the error `CannotDivideByZero()`, it is not reflected in the test

It should be of the form `expect(() => foo()).toThrow(new MyError())`

coxmars:
Hi @zmalatrax I applied the changes and add a few more tests ✅

image

image

src/hints/math/divMod.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    const vm = new VirtualMachine();
+    vm.memory.addSegment();
+    vm.memory.addSegment();
+
+    vm.memory.assertEq(vm.ap, lhsValue);
+    vm.memory.assertEq(vm.ap.add(1), rhsValue);
+
+    expect(() => {
+      divMod(vm, hint.lhs, hint.rhs, hint.quotient, hint.remainder);
+    }).toThrow(CannotDivideByZero);
```
zmalatrax, line 109: We test that the error message are similar in the codebase, using an instance of the error rather than testing that the instance of the class are smilar, see [Jest documentation](https://jestjs.io/docs/expect#tothrowerror) for reference

```suggestion
    }).toThrow(new CannotDivideByZero());
```

coxmars:
Hi @zmalatrax, I already applied the improvement, thanks 🫡

zmalatrax (approved):
good

